PR #3, opened 2017-02-05
add argument --no-path to suppress -gnatef

stephe-ada-guru:
* langkit/libmanage.py (ManageScript.__init__): add argument --no-path
  (ManageScript.gprbuild.run): use --no-path to suppress -gnatef

This allows using Emacs compilation-search-path to find libgprlang/gpr/testsuite/ada/indent.adb (the real source) instead of libgprlang/build/src/indent.adb (a temporary copy) when handling compilation errors.

---
This change is [Reviewable](https://reviewable.io/reviews/adacore/langkit/3)

### DIFF
--- a/langkit/libmanage.py
+++ b/langkit/libmanage.py
@@ -191,6 +191,10 @@ class ManageScript(object):
             '--parsable-errors', '-P', action='store_true', default=False,
             help='Generate error messages parsable by tools'
         )
+        args_parser.add_argument(
+            '--no-path', action='store_true', default=False,
+            help='Do not include path in file names in error messages'
+        )
 
         def create_parser(fn, needs_context=False):
             """
@@ -606,7 +610,7 @@ class ManageScript(object):
             argv.append('-XLIBRARY_TYPE={}'.format(library_type))
             if mains:
                 argv.extend('{}.adb'.format(main) for main in mains)
-            if args.parsable_errors:
+            if args.parsable_errors and not args.no_path:
                 argv.append('-gnatef')
             argv.append('-cargs')
             argv.extend(cargs)


raph-amiard:
Hello Stephen !

I'm not sure about this one, in principle, you're supposed to use --parsable-errors for tool output, in which case you don't care about the absolute paths. If you're using that for your own consumption, then we should rearchitecture the options a bit, rather than have a switch that only has meaning if --parsable-errors is passed.

If you agree I'd propose having:
 * One --classic-errors switch to switch to classic display {file}:{line}:{column}: but with relative paths
 * One --tool-parsable-errors switch that has the current behavior.

What do you think ?

stephe-ada-guru:
> you're supposed to use
> --parsable-errors for tool output, in which case you don't care about
> the absolute paths. 

Mostly. But in Emacs, the compilation buffer shows the full error
messages directly, and it's easier to read without the absolute paths. I
think GPS does the same thing.

The standard Gnu error format
(https://www.gnu.org/prep/standards/standards.html#Errors) is a nice
balance of machine and human readable.

My stated goal in this pull request, of using compilation-path to remap
the actual file location, was not fully successful. Emacs ada-mode uses
the compiler cross-reference information, and it only knows about
build/src/*.adb.

So I'm working on another patch that puts the actual source dir in the
mains.gpr file, rather than copying the source to build/src.

But I'd still like short file names in the error messages.

> If you're using that for your own consumption, then we should
> rearchitecture the options a bit, rather than have a switch that only
> has meaning if --parsable-errors is passed.

We could output full paths in the non-parsable format as well; that
would be useful in a language like Java where the short path can be
ambiguous.

> If you agree I'd propose having:
>  * One --classic-errors switch to switch to classic display
> {file}:{line}:{column}: but with relative paths
>  * One --tool-parsable-errors switch that has the current behavior.
>
> What do you think ?

"classic" can mean full path; the Gnu standard doesn't address this
issue. So to me these two switches mean the same thing.

You said "relative paths"; that usually means relative to some root,
which is not what --no-path means.

For comparison, gnat has the switch --gnatef, which is the inverse of
--no-path; it also only has meaning when the error format is brief.

Now that I think about it, my patch should have had:

 if not args.no_path:
    argv.append('-gnatef')

since 'parseable_errors' has no other effect on the format of errors
emitted by gnat. Perhaps it should affect brief vs verbose error
messages.

pmderodat:
> So I'm working on another patch that puts the actual source dir in the
mains.gpr file, rather than copying the source to build/src.

Please don’t, we are also working on this. :-)

There is something I don’t understand yet: assuming the above is fixed (so you have only one version of a source file at any time), would you still need to get rid of absolute paths?

stephe-ada-guru:
>> So I'm working on another patch that puts the actual source dir in the
>> mains.gpr file, rather than copying the source to build/src.
>
> Please don’t, we are also working on this. :-)

Ok, I won't commit it.

> There is something I don’t understand yet: assuming the above is fixed
> (so you have only one version of a source file at any time), would you
> still need to get rid of absolute paths?

Not _need_, no. But I would prefer short filenames.

pmderodat:
Removing copies in `$BUILD_DIR/src` is now implemented!

Now, I have a proposal to refine your change: what do both of you think of the following: let’s introduce a new option `--diagnostic-style=X` to replace `--parsable-errors`, with `X` being one of:

* `default` (obviously the default ;-)), whose behavior matches the current default (human-readable traceback);
* `gnu-full`, whose behavior matches the current `--parsable-errors`, plus passing -gnatef to GNAT (standard GNU format with full paths);
* `gnu-base`, whose behavior matches `--parsable-errors --no-path` (standard GNU format with base names).

stephe-ada-guru:
Pierre-Marie de Rodat <notifications@github.com> writes:

> Removing copies in `$BUILD_DIR/src` is now implemented!

Excellent; looks exactly like my local patch :).

> Now, I have a proposal to refine your change: what do both of you
> think of the following: let’s introduce a new option
> `--diagnostic-style=X` to replace `--parsable-errors`, with `X` being
> one of:
>
> * `default` (obviously the default ;-)), whose behavior matches the
> current default (human-readable traceback);
> * `gnu-full`, whose behavior matches the current `--parsable-errors`,
> plus passing -gnatef to GNAT (standard GNU format with full paths);
> * `gnu-base`, whose behavior matches `--parsable-errors --no-path`
> (standard GNU format with base names).

looks good to me.

-- 
-- Stephe

raph-amiard:
As stated emoticonically, I agree !

pmderodat:
Hi @stephe-ada-guru,

This new flag is now implemented. :-) Thank you for your help on this, I’m now closing this PR, assuming all is fine for your needs.

PR #9, opened 2017-12-27
Need to install the langkit/utils/ package as well

briot:
Older versions of langkit apparently had a utils.py, but it was now broken into several
python files which need to be installed


pmderodat:
Good catch, thank you @briot! Looking closely at this, I see that we should also add the `gdb` and `stylechecks` subpackages. I’ll do that as well after your change. This also makes me think that we should setup automated testsuite runs that use the `setup.py` facility so that we make sure this does not happen again. Will do!

PR #73, opened 2018-09-03
Work on public structs and copyright headers

pmderodat:


langkit/support/langkit_support-boxes.ads
```diff
@@ -0,0 +1,61 @@
+--  This package provides a simple implementation for boxing types: types that
```
raph-amiard: 1. It would be good to have an action item to migrate this to GNATCOLL.Refcount at some stage
2. I'd rather it has some name that says it is refcounted. Since it's not part of the stdlib you cannot expect ppl to know this terminology IMHO. No strong opinion but there it is :)

PR #76, opened 2018-09-05
DSL: rename LongType, clean up null expressions handling

pmderodat:


langkit/compiled_types.py
```diff
@@ -584,7 +584,7 @@ def is_long_type(self):
 
         :rtype: bool
         """
-        return self == T.LongType
+        return self == T.IntegerType
```
raph-amiard: Meh, I would prefer IntType, so that the type name is not longer afterwards (also sticks better with BoolType for example), after all we're not in Ada so no need to stick for extra verbose type names :)

raph-amiard: Actually thinking about this, I would prefer Int and Bool. the "Type" suffix is just as useless in Python.

PR #92, opened 2018-10-16
Add try_or_else expression to dsl.

Roldak:


langkit/expressions/base.py
```diff
@@ -2590,7 +2591,12 @@ def construct(self):
         from langkit.expressions import Cast
 
         try_expr = construct(self.try_expr)
-        else_expr = construct(self.else_expr)
+
+        if self.else_expr is None:
+            else_expr = NullExpr(try_expr.type)
```
pmderodat: In this case, I think you should first check that it’s legal to use a null expression for the return type. See how we do similar things in Then: https://github.com/AdaCore/langkit/blob/5493449313bd4a7f61591d151fa0790636bb5732/langkit/expressions/boolean.py#L531 Ideally we should factor this out.

PR #108, opened 2018-11-27
Add consistency check option to parse executable

raph-amiard:
To enhance testing of parser in case of error recovery

### DIFF
--- a/langkit/templates/main_parse_ada.mako
+++ b/langkit/templates/main_parse_ada.mako
@@ -1,6 +1,7 @@
 ## vim: filetype=makoada
 
 with Ada.Calendar;              use Ada.Calendar;
+with Ada.Containers.Hashed_Sets;
 with Ada.Containers.Vectors;
 with Ada.Strings;               use Ada.Strings;
 with Ada.Strings.Unbounded;     use Ada.Strings.Unbounded;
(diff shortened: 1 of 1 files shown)


langkit/templates/main_parse_ada.mako
```diff
@@ -199,13 +200,30 @@ procedure Parse is
          Dump_Lexical_Env (Unit);
       end if;
 
+      if Check then
          declare
+            procedure Check_Parents (Node, Parent : ${root_entity.api_name});
```
Roldak: This might not be sufficient to detect duplicated sub-trees. For example, consider the following langkit grammar:
```python
bin_op = Binop(op, Opt("+"), op)
op = List(foo, allow_empty=True)
```
Given an empty string, the parser would generate a Binop with both operands being the same objects (the empty list) but which parents are (rightly) the same.

Roldak: Btw, why is this procedure declared and implemented here?

raph-amiard: Good point about the sub-trees ! Means that I should check for uniqueness rather than parents consistency. This is what I originally wanted to do but then was lazy declaring a set data type ...

PR #124, opened 2019-02-06
S121-011: Revise enum node creation mechanism.

Roldak:


langkit/dsl.py
```diff
@@ -656,186 +795,49 @@ def has_abstract_list(cls):
...
-        def type_ref(self):
-            """
-            TypeRepo.Defer instance for the ASTNodeType subclass corresponding
-            to this alternative.
-            """
-            def get():
-                assert self._type
-                return self._type
-            return T.Defer(get, 'Alternative({})'.format(self.name.camel))
+class EnumNodeAlternative(object):
```
pmderodat: This class is not supposed to be used outside of the Langkit engine (i.e. it’s not part of the DSL), so let’s rename it appropriately (e.g. _EnumNodeAlternative)

langkit/dsl.py
```diff
@@ -517,21 +493,178 @@ def process_subclass(mcs, name, bases, dct, is_root):
...
+            astnode_type = element_type.list
+        else:
+            astnode_type = ASTNodeType(
+                cls._name, cls._location, cls._doc,
+                base=None if is_root else cls._base._resolve(),
+                fields=cls._fields,
+                env_spec=cls._env_spec,
+                annotations=cls._annotations,
+
+                # only enum nodes are abstract at this point
```
pmderodat: Please capitalize “only”

langkit/dsl.py
```diff
@@ -517,21 +493,178 @@ def process_subclass(mcs, name, bases, dct, is_root):
...
+        dct['__new__'] = constructor
+        dct['_create_parser'] = classmethod(
+            create_parser_bool_node if qualifier
+            else create_parser_enum_node
+        )
+
+        dct['_alternatives'] = alts
+        dct['_qualifier'] = qualifier
+
+        # Make Alternative instances available as attributes of the enum node
```
pmderodat: Alternative → _EnumNodeAlternative

langkit/dsl.py
```diff
@@ -517,21 +493,178 @@ def process_subclass(mcs, name, bases, dct, is_root):
...
+        # class for a convenient way to create parsers for them.
```
pmderodat: Thinking out loud: at this point I wonder if there’s much value keeping these instances around. Maybe we should just put parser constructors here (as functions, not classes with __new__ methods) and be done with it. Anyway, that’s for the future :)

PR #130, opened 2019-02-14
Minor refactorings

pmderodat:


langkit/compile_context.py
```diff
@@ -1403,6 +1422,14 @@ def emit(self, file_root='.', main_source_dirs=set(), main_programs=set(),
         :param bool generate_gdb_hook: Whether to generate the
             ".debug_gdb_scripts" section. Good for debugging, but better to
             disable for releases.
+
+        :param list[str] plugin_passes: List of passes to add as plugins to the
+            compilation pass manager. List item must be a name matching the
+            following pattern: ``MODULE.CALLABLE`` where ``MODULE`` is the name
```
raph-amiard: Why u CAP ?

langkit/compile_context.py
```diff
@@ -1605,7 +1605,7 @@ def _compile(self, check_only=False, annotate_fields_types=False):
             PropertyPass('check overriding types',
                          PropertyDef.check_overriding_types),
             PropertyPass('check properties returning node types',
-                         PropertyDef.check_return_types),
+                         PropertyDef.check_returned_nodes),
```
raph-amiard: I think the previous name was better but whatevz

langkit/compile_context.py
pmderodat: I generally use upper case for pattern chunks and lower case for the rest.

langkit/compile_context.py
pmderodat: Before your renaming the previous pass name was “check properties return types” (my intent was “check various things about the return types of properties”), so the method mirorred exactly the name of the pass. I thought your change called for a method rename as well, no?

PR #136, opened 2019-03-01
R628-021: Check for useless dynamic var binds.

Roldak:


testsuite/tests/properties/dynvars/test.out
```diff
@@ -1,4 +1,7 @@
 == abstract: default, concrete: default ==
+File "test.py", line 42, in AbstractNode.use_prop
+  line 42, in bind expression
+    Warning: Useless bind of dynamic var 'env'
```
pmderodat: This warning is irrelevant to the goal of this testcase (which is to test that we ensure that dynamic vars stays consistenc with inheritance). I would suggest to disable this warning in `test.py` rather than letting the warning clutter the test output.

testsuite/tests/properties/invalid_bind/test.out
```diff
@@ -19,6 +19,9 @@ File "test.py", line 31, in BarNode.main_prop
     Error: In Bind's eq_prop BazNode.prop4, some dynamic variables need to be bound: dyn_node
 
 == Correct bind eq_prop A ==
+File "test.py", line 31, in BarNode.main_prop
```
pmderodat: Likewise.

testsuite/tests/properties/useless_bind/test.py
```diff
@@ -0,0 +1,52 @@
...
+                                 langkit_property, ignore)
+from langkit.parsers import Grammar, Or
+
+from utils import emit_and_print_errors
+
+dyn_node = DynamicVariable('dyn_node', T.FooNode)
+dyn_node_2 = DynamicVariable('dyn_node_2', T.FooNode)
+
+class FooNode(ASTNode):
+    @langkit_property(public=True)
```
pmderodat: For testcase readability, I would suggest to rename the various properties depending on what they are used for. For instance: `test_propX` for the properties that constitute the warning test, and `helperX` for properties that exist only to provide material for tests.

PR #139, opened 2019-03-07
S213-021: Work on exceptions in bindings

Roldak:


pmderodat:
For the record: ok, but missing so much doc! Unacceptable tableflip!

PR #142, opened 2019-03-12
S130-035: Move preconditions from public spec to internal implem & co.

Roldak:


langkit/documentation.py
```diff
@@ -196,6 +196,10 @@ def instantiate_templates(doc_dict):
         Nevertheless, the library does its best not to crash the program,
         materializing internal errors using this kind of exception.
     """,
+    'langkit.precondition_failure': """
+        Exception raised when user-input passed to a langkit-provided API did
+        not satisfy the specification of that API.
```
pmderodat: I suggest we avoid referring to Langkit in generated APIs (i.e. when talking to the users of generated libraries). What about:
```
Exception raised when an API is called while its preconditions are not satisfied.
```

langkit/templates/pkg_rewriting_impl_body_ada.mako
```diff
@@ -27,11 +27,138 @@ package body ${ada_lib_name}.Rewriting_Implementation is
      (Rewriting_Handle_Pointer, Rewriting_Handle);
    pragma Warnings (On, "possible aliasing problem for type");
 
+   --------------
+   -- PRE_FAIL --
+   --------------
```
pmderodat: Please use our standard coding convention: case-preserving comment boxes, one comment box per subprogram and the documentation with the subprogram declaration (not body, so add declarations for all subprogram bodies). Thanks!

pmderodat: The commit right after this one removes these subprograms… so at this point, why not merging the two commits to do the right thing from the start?

langkit/templates/pkg_rewriting_impl_body_ada.mako
```diff
@@ -27,136 +27,120 @@ package body ${ada_lib_name}.Rewriting_Implementation is
...
+      if not Value then
+         raise Precondition_Failure with Msg;
       end if;
-   end Pre_Check_Null_Or_Valid_Context;
-   -- Check that the node rewriting handle is either null, or belongs to the
-   -- given rewriting handle.
+   end Pre_Check;
+
+   <%def name="pre_check_rw_handle(handle)">
+      --  Check that rewriting handle ${handle} is not null
```
pmderodat: These comments document the macro itself, but will clutter the generated code. Can you turn them into Mako comments? (## Check that […])

langkit/templates/pkg_rewriting_impl_body_ada.mako
```diff
@@ -27,136 +27,120 @@ package body ${ada_lib_name}.Rewriting_Implementation is
...
+   </%def>
+
+   <%def name="pre_check_child_index(handle, index, for_insertion)">
+      <%
+      upper_bound = "Children_Count ({}){}".format(
+         handle,
+         " + 1" if for_insertion else ""
+      )
+      %>
+      --  Check that ${index} is in range 1 .. ${upper_bound}.
```
pmderodat: Like the above, but also document the other arguments in the Mako comment (index, for_insertion).

langkit/templates/pkg_rewriting_impl_body_ada.mako
```diff
@@ -27,136 +27,120 @@ package body ${ada_lib_name}.Rewriting_Implementation is
...
+      Pre_Check
+        (${index} <= ${upper_bound},
+         "Invalid index " & ${index}'Image & ": ${handle} has " &
+         Children_Count (${handle})'Image & " children");
```
pmderodat: This string formatting will happen everytime the corresponding check is performed, looks wasteful. I suggest to remove the dynamic computation in order to have a static string: `Out of bound child index` seems enough for the caller to understand what’s wrong.

Or even better: inline Pre_Check, so that the string formatting happens iff there is a precondition failure.

langkit/templates/pkg_rewriting_impl_body_ada.mako
```diff
@@ -188,46 +315,55 @@ package body ${ada_lib_name}.Rewriting_Implementation is
...
-      Context        : constant Internal_Context := Unit.Context;
-      Context_Handle : constant Rewriting_Handle := Handle (Context);
-      Filename       : constant Unbounded_String :=
-         To_Unbounded_String (Get_Filename (Unit));
-      Cur            : constant Cursor := Context_Handle.Units.Find (Filename);
    begin
-      if Cur /= No_Element then
-         return Element (Cur);
-      end if;
+      ${pre_check_rw_handle("Handle (Context (Unit))")}
```
pmderodat: The check that `Unit` should not have diagnostics is missing.

langkit/templates/pkg_rewriting_impl_body_ada.mako
```diff
@@ -27,11 +27,132 @@ package body ${ada_lib_name}.Rewriting_Implementation is
...
+         "${node} should be associated to rewriting context ${handle}.");
+   </%def>
+
    function Handle (Context : Internal_Context) return Rewriting_Handle is
      (Convert (Get_Rewriting_Handle (Context)));
 
    function Context (Handle : Rewriting_Handle) return Internal_Context is
-     (Handle.Context);
+   begin
+      ${pre_check_rw_handle("Handle")}
```
pmderodat: Except in `<%def>` directives, please replace all double quotes with single ones (I try to keep this use consistent unless for string literals that must contain single quotes).

langkit/templates/pkg_rewriting_impl_body_ada.mako
```diff
@@ -569,9 +716,9 @@ package body ${ada_lib_name}.Rewriting_Implementation is
    -- Kind --
    ----------
 
-   function Kind (Handle : Node_Rewriting_Handle) return ${root_node_kind_name}
-   is
+   function Kind (Handle : Node_Rewriting_Handle) return ${root_node_kind_name} is
```
pmderodat: Line seems to exceed 80 columns. Please run the testsuite’s style checkers!

langkit/templates/pkg_rewriting_impl_body_ada.mako
```diff
@@ -627,6 +780,9 @@ package body ${ada_lib_name}.Rewriting_Implementation is
       Child  : Node_Rewriting_Handle)
    is
    begin
+      ${pre_check_nrw_handle("Handle")}
+      ${pre_check_child_index("Handle", "Index", False)}
```
pmderodat: The check that `Child` must not be tied is missing.

langkit/templates/pkg_rewriting_impl_body_ada.mako
```diff
@@ -715,6 +881,11 @@ package body ${ada_lib_name}.Rewriting_Implementation is
       Index  : Positive;
       Child  : Node_Rewriting_Handle) is
    begin
+      ${pre_check_nrw_handle("Handle")}
+      ${pre_check_is_list_kind("Kind (Handle)")}
+      ${pre_check_child_index("Handle", "Index", True)}
```
pmderodat: Please add “for_insertion=” (so we have a name, and thus visible meaning, near this strange looking bool literal).

langkit/templates/pkg_rewriting_impl_body_ada.mako
```diff
@@ -841,22 +1027,30 @@ package body ${ada_lib_name}.Rewriting_Implementation is
...
-      Result : Node_Rewriting_Handle := Allocate
-        (Kind, Handle, No_Unit_Rewriting_Handle, No_Node_Rewriting_Handle);
+      Children : Node_Rewriting_Handle_Array) return Node_Rewriting_Handle is
    begin
-      Result.Children := (Kind   => Expanded_Regular,
-                          Vector => <>);
-      Result.Children.Vector.Reserve_Capacity (Children'Length);
+      ${pre_check_rw_handle("Handle")}
+      ${pre_check_is_not_token_kind("Kind")}
       for C of Children loop
```
pmderodat: Given that `C` will appear in the error message, please rename it so something more explicit, like `One_Child`.

Likewise for `A` in Create_From_Template.

langkit/documentation.py
```diff
@@ -743,9 +743,11 @@ def instantiate_templates(doc_dict):
         Set the root node for the unit Handle to Root. This unties the previous
         root handle. If Root is not No_Node_Rewriting_Handle, this also ties
         Root to Handle.
+        Root must not already be tied to another analysis unit handle.
```
pmderodat: Please add an empty line to separate paragraphs (otherwise paragraphs will be merged during template rendering). Here and below, whenever you want to create a new paragraph.

langkit/documentation.py
```diff
@@ -791,14 +793,19 @@ def instantiate_templates(doc_dict):
         If Handle is the root of an analysis unit, untie it and set New_Node as
         its new root. Otherwise, replace Handle with New_Node in Handle's
         parent node.
+        Note that:
+         - Handle must be tied to an existing analysis unit handle.
```
pmderodat: Comments are interpreted as Sphinx, so please remove indentation for lists. Also for consistency, I would recommend to replace dashes (`-`) with stars (`*`). Thanks!

PR #157, opened 2019-03-29
S328-024: Compute optional field information.

Roldak:


langkit/compile_context.py
```diff
@@ -882,6 +882,27 @@ def compute_types(self):
...
+            for field in node_type.get_parse_fields(include_inherited=False)
+        ]
+
+        for field in all_parse_fields:
+            field._is_optional = False
+            for parser in field.parsers_from_transform:
+                # If parser is an Opt parser and is not set to produce an enum
+                # alternative, it means that field is optional.
+                if isinstance(parser, Opt) and not parser._booleanize:
+                    field._is_optional = True
```
pmderodat: If I understand correctly, the goal is not really to determine if an `Opt` parser is involved, but rather if the parser can return a null node, right? If so, I think this analysis should be much more involved. For instance when you have an `Or` subparser, you then need to check if any subparser for this `Or` can itself return a null node. You must also check for `Null` parsers, recurse in `Defer` parsers, etc.

PR #158, opened 2019-03-31
Support static-pic

anisimkov:


pmderodat:
Tracking offline discussions:

We don’t want to force the build of static-pic for everyone who want to build in static mode, but only to allow one more build mode.

This is also the opportunity to revamp the --enable/disable-shared/static flags, so we’ll provide another change to implement this.

PR #161, opened 2019-04-02
Support force install

anisimkov:


langkit/libmanage.py
```diff
@@ -257,6 +257,11 @@ def internal(*args, **kwargs):
             help='Installation directory.'
         )
 
+        install_parser.add_argument(
+            '--force', '-f', action='store_true', dest='force', default=False,
```
pmderodat: Both `dest` and `default` arguments can be removed (`default=False` is implied by `action='store_true'` and `dest='force'` is implied by `--force`).

anisimkov: ok, good, i'll fix

anisimkov: Fixed

pmderodat: Thanks!

PR #162, opened 2019-04-02
libmanage.py: rework options to select library types

pmderodat:


pmderodat:
> library types list in help message wrong:
> help='Comma-separated list of library types to build (relocatable,'
> ' static-pic and pic). By default, build only shared'
> ' libraries.'

Good finding, thanks! I’ll fix that.

> Are the parentheses around entire boolean exception mandatory?

Yes, otherwise you’ll get a syntax error (you can try ;-))

PR #166, opened 2019-04-04
Work on the introspection API

pmderodat:


langkit/templates/pkg_introspection_body_ada.mako
```diff
@@ -268,27 +346,94 @@ package body ${ada_lib_name}.Introspection is
...
+      % else:
+         ${return_program_error()}
       % endif
+   end Fields;
 
+   ------------
+   -- Fields --
+   ------------
+
+   function Fields
```
raph-amiard: This looks like expensive work that could be cached no ?

langkit/templates/pkg_introspection_body_ada.mako
```diff
@@ -9,11 +9,38 @@ with Ada.Strings.Unbounded.Hash;
 with ${ada_lib_name}.Implementation;
```
raph-amiard: General comment: this commit seems to be a bit doc-lacking. There are both undocumented types and IMHO the code generation part could be a bit more documented.

langkit/templates/pkg_introspection_spec_ada.mako
```diff
@@ -68,21 +68,37 @@ package ${ada_lib_name}.Introspection is
    function Derived_Types (Id : Node_Type_Id) return Node_Type_Id_Array;
    --  Return type references for all direct derivations for Id
 
+   <% all_abstract = ctx.sorted_parse_fields + ctx.sorted_properties %>
+
+   type Abstract_Node_Data_Reference is
```
raph-amiard: While this is the Langkit name, maybe `Node_Data_Reference` is sufficient for the public API name, what do you think ?

langkit/templates/pkg_introspection_spec_ada.mako
```diff
@@ -68,6 +68,12 @@ package ${ada_lib_name}.Introspection is
    function Derived_Types (Id : Node_Type_Id) return Node_Type_Id_Array;
    --  Return type references for all direct derivations for Id
 
+   function Is_Derived_From (Id, Parent : Node_Type_Id) return Boolean;
+   --  Return whether true if at least one of the following is true:
```
raph-amiard: This comment made my head hurt. This is the definition of the function but including the recursive definition in the comment is not very helpful IMHO :) Saying it returns true if the type identified by Id is either equal to parent, directly or indirectly derived from parent would seem better

langkit/templates/pkg_introspection_body_ada.mako
pmderodat: As discussed live: this algorithm runs in linear time for number of fields + number of derivations. Caching is possible, but I would suggest to wait for reported performance issue to bother introducing a cache. :-)

langkit/templates/pkg_introspection_body_ada.mako
pmderodat: After live discussion: this note was written thinking this patch hunk was for a package specification (which contains most documentation). This is the body, so the API is not documented here on purpose, but internals should be discribed as needed. We’ll add more comments if necessary.

langkit/templates/pkg_introspection_spec_ada.mako
pmderodat: Good point, I’ll rename, thanks!

langkit/templates/pkg_introspection_spec_ada.mako
pmderodat: Right, sorry, the comment describes the computation more than what it expresses. I’ll fix that.

PR #168, opened 2019-04-05
Add generation of OCaml bindings

danielmercier:
OCaml bindings generation from langkit.

For now, there are no tests, since the generated OCaml API might change.

The idea is to let you begin the reviewing process and give feedback on it.


langkit/expressions/base.py
```diff
@@ -1627,6 +1636,15 @@ def render_introspection_constant(self):
         """
         raise not_implemented_error(self, self.render_introspection_constant)
 
+    def render_ocaml_constant(self):
+        """
+        assuming this expression is a valid constant, return ocaml code to
```
raph-amiard: Same comment

langkit/expressions/base.py
```diff
@@ -1644,6 +1662,9 @@ def render_public_ada_constant(self):
     def render_python_constant(self):
         return str(self.value)
 
+    def render_ocaml_constant(self):
+        return "true" if self.value else "false"
```
raph-amiard: `str(self.value).lower()` ?

langkit/expressions/base.py
```diff
@@ -1690,6 +1714,9 @@ def render_public_ada_constant(self):
     def render_python_constant(self):
         return repr(self.value)
 
+    def render_ocaml_constant(self):
+        return '(Char.code {})'.format(ord(self.value))
```
raph-amiard: Curious about why you need this ? Should probably document why code equivalent to the python one isn't sufficient

langkit/ocaml_api.py
```diff
@@ -0,0 +1,444 @@
...
+            (T.Symbol, lambda _: False),
+            (ct.EnumType, lambda t: False),
+            (ct.ASTNodeType, lambda t:
+                len(t.concrete_subclasses) == 0),
+            (ct.EntityType, lambda t: self.is_empty_type(t.astnode)),
+            (T.AnalysisUnit, lambda _: False),
+            (ct.ArrayType, lambda t: self.is_empty_type(t.element_type)),
+            (ct.StructType, lambda t:
+                len([
+                    t
```
raph-amiard: It looks like you want to check if there is no field that has an empty type ? I'd recommend in that case writing:

```python
not any(self.is_empty_type(f.type) for f in t.get_fields())
```

Also nit: strange formatting, I'd write

```python
                    t for field in t.get_fields()
                    if not self.is_empty_type(field.type)
```

langkit/ocaml_api.py
```diff
@@ -0,0 +1,444 @@
...
+        :rtype: list[ct.CompiledType]
+        """
+        if for_module == from_module:
+            return 'fields'
+        else:
+            return "{}.{}".format(self.module_name(for_module), 'fields')
+
+    def check_for_null(self, c_value, astnode, context):
+        """
+        Return an ocaml expression that first check if the underlying node of
```
raph-amiard: "that first checks"

langkit/ocaml_api.py
```diff
@@ -0,0 +1,444 @@
...
+
+    def get_properties(self, tpe):
+        """
+        Return the list of all properties that should be exposed to the user.
+
+        :rtype: list[ct.BaseField]
+        """
+
+        return [
+            field
```
raph-amiard: same comment about list comp formatting

langkit/ocaml_api.py
```diff
@@ -0,0 +1,444 @@
...
+            for field in tpe.fields_with_accessors()
+            if field.is_property
+        ]
+
+    def array_wrapper(self, array_type):
+        return (ct.T.entity.array
+                if array_type.element_type.is_entity_type else
```
raph-amiard: Being nitty again :) 

```python
ct.T.entity.array
if array_type.element_type.is_entity_type
else array_type
```

langkit/ocaml_api.py
```diff
@@ -0,0 +1,444 @@
...
+            (T.AnalysisUnit, lambda t: t.api_name.camel),
+            (ct.ArrayType, lambda t: t.api_name.camel),
+            (ct.StructType, lambda t: t.api_name.camel),
+            (T.BigInt, lambda t: 'BigInteger'),
+        ])
+
+    def wrap_requires_context(self, type):
+        """
+        Returns true if the given type need the context to be wrapped in an
+        ocaml value.
```
raph-amiard: Casing of "OCaml" is inconsistent. Pick one ;)

langkit/ocaml_api.py
```diff
@@ -0,0 +1,444 @@
...
+            (ct.EntityType, lambda _: True),
+            (T.Token, lambda _: False),
+            (T.Symbol, lambda t: False),
+            (T.Bool, lambda _: False),
+            (T.Int, lambda _: False),
+            (T.Character, lambda _: False),
+            (ct.ArrayType, lambda t:
+                self.wrap_requires_context(t.element_type)),
+            (ct.StructType, lambda t:
+                len([t
```
raph-amiard:
```python
any(self.wrap_requires_context(f.type) for f in t.get_fields())
```

langkit/templates/ocaml_api/module_ocaml.mako
```diff
@@ -0,0 +1,1174 @@
...
+    let uchar = UText.get text 0 in
+    Unsigned.UInt32.of_int (UChar.code uchar)
+
+  let c_type = view uint32_t ~read:wrap ~write:unwrap
+end
+
+module BigInteger = struct
+
+  (* TODO: use a int for now, should switch to a big integer. *)
+  type t = int
```
raph-amiard: ^^

langkit/expressions/base.py
danielmercier: Actually, this is pretty tricky, and is false for now, good catch.
I guess that characters can be unicode?
In OCaml bindings, characters are represented as utf-8 strings.
So I guess that I should also make the convertion here

pmderodat: Yes, in Langkit, strings are represented as arrays of wide wide characters (i.e. full Unicode codepoints). If the idiomatic type to use in OCaml is UTF-8 strings, yes: do the conversion each time a value goes from our C API through the OCaml binding and conversely.

langkit/ocaml_api.py
pmderodat: I concur. By the way, could you enhance the docstring to describe what an empty type is exactly? Looking at the implementation, it looks like it will return False for every type.

langkit/ocaml_api.py
```diff
@@ -0,0 +1,444 @@
...
+            return "{}.{}".format(self.module_name(for_module), 'fields')
+
+    def check_for_null(self, c_value, astnode, context):
+        """
+        Return an ocaml expression that first check if the underlying node of
+        c_value is null. If it is the case, the expression is evaluated to
+        None, otherwise, wrap c_value to the given astnode type.
+
+        :param str c_value: The ocaml c value for a node.
+        :param ct.ASTNodeType: The type of c_value.
```
pmderodat: You forgot the name of the parameter here

langkit/ocaml_api.py
```diff
@@ -0,0 +1,444 @@
...
+        :param str context: The ocaml expression denoting the context.
+        """
+
+        return (
+            "if is_null (getf {} EntityStruct.node) then None else Some ({})".
+            format(c_value, self.wrap_value(c_value, astnode, context))
```
pmderodat: In this case we generally put the dot at the beginning of the next line: could you move it?

langkit/ocaml_api.py
danielmercier: I will update the comment. By "empty type" I mean empty struct, or struct that only contains empty structs.
In the implementation, the function returns true when a struct type does not have any field, or the only fields that it has are also empty structs.

langkit/ocaml_api.py
```diff
@@ -0,0 +1,444 @@
...
+        if len(super_type.concrete_subclasses) == len(concrete_precise_types):
+            concrete_precise_types = [super_type]
+
+        return concrete_precise_types
+
+    def get_parse_fields(self, tpe):
+        """
+        Return all the parse fields to be exposed as a field in record of tpe.
+
+        :param ct.ASTNodeType tpe: The ast node type of which we want to get
```
pmderodat: We haven’t renamed all the identifers yet, but for new names/docs we try to avoid mentioning “AST” since the tree that Langkit handles is a concrete syntax tree. So I suggest rewording to just “The node of which […]”. Also we generally don’t abbreviate `tpe` this way :-) I would suggest to rename it to `node` as it’s even more specific than any type.

pmderodat: (Same comment for other occurences)

langkit/ocaml_api.py
pmderodat: Hm… we haven’t written it out in our coding style, but for Python IIRC we agreed several years ago to use the formatting that Daniel used. Anyway, looks like we’re not very consistent on this point.

langkit/ocaml_api.py
pmderodat: I vote for OCaml as soon as we designate the language :-) (not the interpreter program)

langkit/ocaml_api.py
pmderodat: Interesting… Does this take into account inherited fields? (for nodes, which are also structs) If nodes with only inherited fields are not empty, then we probably never had any empty struct. I guess it’s allowed, but since it’s never been tested I would not bet it works. ;-)

danielmercier: nodes are handled separately and are considered empty if they do not have any concrete subclasses. Even if a node has no field, we still have the c_value so the OCaml struct is not empty. (I might have misunderstood your concern)

pmderodat: No, you’re fine. :-)

The case you handle here (structs with no field and abstract nodes with no concrete subclass) seem quite pathological, and I’m tempted to make sure they never happen in practice (i.e. make them illegal for Langkit). That would enable you to simplify your change. What do you think?

danielmercier: Yes, I had to make this because it happened in some of langkit's tests.
One example is the Metadata structure that sometimes does not contain any field.
I also saw some ASTNode without any subclasses in the testsuite.

It would definitely simplify some code since I would not need to check whether an empty OCaml type would be generated.

pmderodat: Oh right, I forgot about metadata: I’m taking this back (we’ll keep supporting empty structs). And I just remembered that we need to support abstract classes with no concrete subclass because of the way our generic list type works. So never mind :-) Thank you for updating the docstring, then!

Note that you could simplify the call to dispatch_on_type: for instance replacing all cases that unconditionally return False with a single `(ct.CompiledType, lambda _: False)` clause at the end of the list.

PR #169, opened 2019-04-08
Add support for properties in the introspection API

pmderodat:


langkit/compiled_types.py
```diff
@@ -754,6 +758,26 @@ def memoization_kind(self):
         """
         return (names.Name('Mmz') + self.name).camel_with_underscores
 
+    @property
+    def introspection_radix(self):
```
raph-amiard: introspection_root or introspection_root_name would sound more familiar to me :)

langkit/templates/pkg_introspection_spec_ada.mako
```diff
@@ -177,4 +285,91 @@ package ${ada_lib_name}.Introspection is
    --  exception.
    % endif
 
+private
+
+   type Value_Record;
+   type Value_Access is access all Value_Record;
+
+   --  In order to avoid Any_Value_Type to be tagged (which makes all its
```
raph-amiard: Ewwww

raph-amiard: Ada :'(

langkit/templates/pkg_introspection_spec_ada.mako
```diff
@@ -239,6 +239,13 @@ package ${ada_lib_name}.Introspection is
    function Field_Type (Field : Field_Reference) return Node_Type_Id;
    --  Return a reference to the node type that covers what Field can contain
 
+   function Evaluate_Field
```
raph-amiard: `Eval_Field`/`Get_Field`, shorter and just as clear

langkit/templates/pkg_introspection_spec_ada.mako
```diff
@@ -295,6 +295,17 @@ package ${ada_lib_name}.Introspection is
    --  Return the lower-cased name for Property's argument whose index is
    --  Argument_Number.
 
+   function Evaluate_Property
```
raph-amiard: Same comment, although makes me think that `Eval_Property` is better than `Get_Property` which makes no sense

PR #175, opened 2019-04-24
S418-018: Refactor add_to_env.

Roldak:
- "env_assoc"s now take themselves the dest_env and the metadata info.
- Add add_to_env_kv and new_env_assoc to simplify common usage.

### DIFF
--- a/testsuite/tests/lexical_envs/add_to_env_mult_dests/main.py
+++ b/testsuite/tests/lexical_envs/add_to_env_mult_dests/main.py
@@ -1,0 +1,22 @@
+from __future__ import absolute_import, division, print_function
+
+print('main.py: Running...')
+
+
+import sys
+
+import libfoolang
+
+
+ctx = libfoolang.AnalysisContext()
+u = ctx.get_from_file('input')
+if u.diagnostics:
+    for d in u.diagnostics:
+        print(d)
+    sys.exit(1)
+
+u.populate_lexical_env()
+for i in u.root.findall(libfoolang.Id):
+    print("Scope for {}: {}".format(i, i.p_get_scope))
+
+print('main.py: Done.')
--- a/testsuite/tests/lexical_envs/add_to_env_mult_dests/test.py
+++ b/testsuite/tests/lexical_envs/add_to_env_mult_dests/test.py
@@ -1,0 +1,71 @@
+"""
+Test add_to_env with multiple env_assoc having different destination
+environments.
+"""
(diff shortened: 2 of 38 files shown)


testsuite/tests/lexical_envs/add_to_env_mult_dests/test.py
```diff
@@ -0,0 +1,71 @@
+"""
+Test that uses an add_to_env actions with multiple env_assoc having different
+destination environments.
```
pmderodat: I think this needs rewording ;-)

testsuite/tests/lexical_envs/add_to_env_mult_dests/main.py
```diff
@@ -0,0 +1,27 @@
...
+    sys.exit(1)
+
+try:
+    u.populate_lexical_env()
+    for i in u.root.findall(libfoolang.Id):
+        print("Scope for {}: {}".format(i, i.p_get_scope))
+except libfoolang.PropertyError as exc:
+    print('   Got an exception: {}'.format(str(exc).strip()))
+else:
+    print('   All good...')
```
pmderodat: Given that you don’t except any exception here, the except/else clauses could be removed (and thus keep only the content of the try clause)

testsuite/tests/lexical_envs/add_to_env_mult_dests/test.py
Roldak: Why so? (except that `actions` should be `action`)

testsuite/tests/lexical_envs/add_to_env_mult_dests/main.py
Roldak: Right!

testsuite/tests/lexical_envs/add_to_env_mult_dests/test.py
pmderodat: Ah sorry, I did read “Test” as a verb rather than a name, so I had trouble parsing this sentense. IIRC we generally document tests like functions (i.e. with imperative tense). Would you be okay with rewording this way?

testsuite/tests/lexical_envs/add_to_env_mult_dests/main.py
pmderodat: Thanks! I’ll let you adjust this PR: LGTM otherwise.

PR #181, opened 2019-05-24
Various improvements and fixes

pmderodat:


langkit/templates/pkg_implementation_body_ada.mako
```diff
@@ -1296,12 +1296,18 @@ package body ${ada_lib_name}.Implementation is
    function Text
      (Node : access ${root_node_value_type}'Class) return Text_Type is
    begin
+      --  No text is associated to synthetic and ghost nodes
+
       % if ctx.synthetic_nodes:
```
raph-amiard: Seeing this code, it makes me wonder if you'd not rather generate a `Synthetic_Nodes` subtype for which the predicate is just `False` when there are no synthetic nodes. Two reasons spring to mind:

1. First it would simplify this kind of code
2. Second it would go in the direction of being able to treat langkit generated libraries 'uniformly' from an user point of view. So you'd rather have a synthetic_nodes subtype that is an empty set, rather than not having it.

langkit/templates/pkg_common_spec_ada.mako
```diff
@@ -91,6 +91,15 @@ package ${ada_lib_name}.Common is
       % endif
    % endfor
 
+   ## Output a subtype to materialize the set of kinds for synthetic nodes
+   % if ctx.synthetic_nodes:
+      subtype Synthetic_Nodes is ${root_node_kind_name}
+         with Dynamic_Predicate =>
```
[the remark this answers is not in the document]
raph-amiard: Thinking about it, having a subtype is good, but having a function (could be a built-in property to simplify implementation and binding) on nodes in addition seems even better.

Thinking about the day where one would be able to treat nodes from LK libraries uniformly, that would allow checking for syntheticness in a language agnostic way.

langkit/expressions/base.py
```diff
@@ -2923,6 +2923,7 @@ def render(*args, **kwargs):
         assign_var=assign_var,
```
raph-amiard: Yeyyyyyyyyyyyyyyyyyyyy :D

langkit/templates/properties/def_ada.mako
```diff
@@ -29,6 +29,10 @@ is
    Self : ${Self.type.name} := ${Self.type.name}
      (${property.self_arg_name});
 
+   ## Dispatchers must not memoize: it is the job of the static properties do
```
raph-amiard: "Dispatchers must not memoize because it will be done at the static property level, so we don't want to do it twice" seems clearer/more accurate.

langkit/compiled_types.py
```diff
@@ -2404,6 +2419,42 @@ def __init__(self, name, location, doc, base, fields,
...
+            precise_types = (self.precise_list_element_types
+                             .minimal_matched_types)
+            if len(precise_types) > 1:
+                result_list = result.splitlines() if result else []
+                type_descr = [
+                    'This list node can contain one of the following nodes:'
+                ] + list(sorted('* {}'.format(t.dsl_name)
+                                for t in precise_types))
+
+                # Indent lines in `type_descr` as much as the first non-null
```
raph-amiard: Wouldn't a bit of mako have helped there ? this is particularly hard to read :/

langkit/compiled_types.py
```diff
@@ -1769,46 +1788,85 @@ def is_optional(self):
...
+                    extra_lines)
+
+        precise_types = self.precise_types.minimal_matched_types
+
+        # If the field always contains a list, try to give precise types
+        # for the list items.
+        if len(precise_types) == 1 and self.type.is_list_type:
+            precise_element_types = (self.precise_element_types
+                                     .minimal_matched_types)
+            if len(precise_element_types) > 1:
```
raph-amiard: Same comment about makoification

langkit/compiled_types.py
pmderodat: As discussed on Slack: unfortunately the result is piped to Sphinx, which is particularly picky about formatting aspects such as indentation. Using Mako in this context would probably make things worse. That being said, I could probably improve readability of this moving text formatting primitives to `langkit.utils` for instance. I’ll see what I can do.

langkit/templates/pkg_common_spec_ada.mako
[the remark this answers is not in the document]
pmderodat: Good point, I’ll add this, thanks!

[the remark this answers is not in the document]
pmderodat: Actually this made me realize that we lacked a is-synthetic predicate in the C/Python APIs, so good catch!

langkit/templates/pkg_implementation_body_ada.mako
pmderodat: Good idea, will do!

PR #185, opened 2019-06-02
Fix structure of arrays

raph-amiard:


pmderodat:
d5307d8 introduces a syntax error. I’ve reorganized the commits and opened #186.

PR #198, opened 2019-07-10
Fix Node.Node_Env

raph-amiard:
Previous version was overcomplicated and didn't generate proper code for
a concrete subclass of a concrete base class.

### DIFF
--- a/langkit/compiled_types.py
+++ b/langkit/compiled_types.py
@@ -2513,6 +2513,16 @@ class ASTNodeType(BaseStructType):
         del entity_type
 
     @property
+    def effective_env_spec(self):
+        """
+        Return the env spec, for this node, whether it's defined on this node
+        or inherited from a parent node.
+        """
+        return self.env_spec if self.env_spec else (
+            self.base.effective_env_spec if self.base else None
+        )
+
+    @property
     def doc(self):
         result = super(ASTNodeType, self).doc
 
--- a/langkit/compiled_types.py
+++ b/langkit/compiled_types.py
@@ -2407,7 +2407,6 @@ class ASTNodeType(BaseStructType):
 
         if env_spec:
             env_spec.ast_node = self
-        self.is_env_spec_inherited = env_spec is None
 
         self.env_spec = env_spec
         """
--- a/langkit/templates/pkg_implementation_body_ada.mako
+++ b/langkit/templates/pkg_implementation_body_ada.mako
@@ -1256,8 +1256,7 @@ package body ${ada_lib_name}.Implementation is
    is
    begin
 
-      <%self:case_dispatch
-        pred="${lambda n: n.env_spec and not n.is_env_spec_inherited}">
+      <%self:case_dispatch pred="${lambda n: n.env_spec}">
       <%def name="action(node)">
          return ${node.name}_Pre_Env_Actions
            (${node.name} (Self), Bound_Env, Root_Env, Add_To_Env_Only);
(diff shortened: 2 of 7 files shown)


langkit/templates/pkg_implementation_body_ada.mako
```diff
@@ -2795,20 +2775,17 @@ package body ${ada_lib_name}.Implementation is
...
-               return Get_Parent_Env;
+                  return Node.Self_Env;
             % endif
-
-         % else:
-            return Node.Self_Env;
-         % endif
+         </%def>
+         <%def name="default()"> return Node.Self_Env; </%def>
+         </%self:case_dispatch>
```
pmderodat: This will generate one `when` clause for each node that has an effective env spec. So lots of redundant code. What about instead computing the list of kinds for which this must return `Get_Parent_Env` and generate at most two clauses instead? (like the previous version did)

langkit/compiled_types.py
```diff
@@ -2512,6 +2512,16 @@ def __init__(self, name, location, doc, base, fields,
         entity_type = self.entity
         del entity_type
 
+    @property
+    def inherited_env_spec(self):
+        """
+        Return the env spec, for this node, whether its defined on this node or
```
pmderodat: Typo (“its” should be “it’s”)

langkit/compiled_types.py
```diff
@@ -2512,6 +2512,16 @@ def __init__(self, name, location, doc, base, fields,
         entity_type = self.entity
         del entity_type
 
+    @property
+    def inherited_env_spec(self):
```
pmderodat: This name is misleading, as the result is not always the inherited spec. What about `effective_env_spec` instead?

langkit/templates/pkg_implementation_body_ada.mako
raph-amiard: Well the generated code might be redundant, but the template is made much simpler, and the resulting machine code will very probably be just as simple. Given the number of time we have gotten this function wrong, I'd rather keep this simple and readable (like the previous version wasn't IMHO).

langkit/compiled_types.py
raph-amiard: Thanks, will fix

langkit/compiled_types.py
raph-amiard: Right, I agree about that. I don't much like `effective_env_spec` either. I think the proper fix would be to call the property `env_spec`, and rename the field something like `explicit_env_spec`. But that's a bigger refactoring.

langkit/templates/pkg_implementation_body_ada.mako
pmderodat: But that’s unrelated: the previous version was wrong because its author (me, IIRC) did not reason correctly about what it should do, yielding a complex and wrong algorithm in the process. I think generating a simple `if` check instead of a big `case` block won’t make the above more complex.

langkit/compiled_types.py
pmderodat: Fair enough about the bigger refactoring. Can you elaborate why you don’t like `effective_env_spec`?

langkit/templates/pkg_implementation_body_ada.mako
raph-amiard: Well it implies re-encoding some of the logic about iterating on nodes and collapsing branches that is not trivial, and is automated by the `case..` block. But probably not undoable indeed.

PR #201, opened 2019-07-17
S308-026 make binaries that depend on ocaml bindings relocatable

yakobowski:
The path to Camomile's standard library, determined _at compile time_
is stored in CamomileLibrary.DefaultConfig, and used at execution
time. As a result, binaries that use Camomile in a naive way
(including our Libadalang OCaml bindings) are not relocatable. Fix this.

### DIFF
--- a/langkit/templates/ocaml_api/module_ocaml.mako
+++ b/langkit/templates/ocaml_api/module_ocaml.mako
@@ -90,7 +90,39 @@ let initialize = foreign ~from:c_lib "${capi.lib_name}_initialize"
 let () = initialize ()
 
 (* Module used to encode/decode UTF32 strings *)
-module Camomile = CamomileLibrary.Make (CamomileLibrary.DefaultConfig)
+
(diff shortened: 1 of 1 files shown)


langkit/templates/ocaml_api/module_ocaml.mako
```diff
@@ -90,7 +90,39 @@ let initialize = foreign ~from:c_lib "${capi.lib_name}_initialize"
...
+   - otherwise we fail
+*)
+
+module CamomileDefaultConfig = CamomileLibrary.DefaultConfig
+
+module CamomileShareConfig = struct
+  let (^/) = Filename.concat
+  let bin_dir = Filename.dirname Sys.executable_name
+  let share_dir = bin_dir ^/ Filename.parent_dir_name ^/ "share" ^/ "camomile"
+  let datadir    = Filename.concat share_dir "database"
```
danielmercier: You can use the operator here too

PR #202, opened 2019-07-25
S701-037: Fix invalidation of env_getter cache.

Roldak:
The Is_Stale procedure was missing information to determine if rebindings
from a given rebound_env where still valid or not and would unsoundly
consider them valid all the time.

This fix adds a field to rebound environments indicating the version of
the context at the time the rebound env is created. Is_Stale now compares
this number with the current context version to determine if a cached
rebound_env is stale or not.

### DIFF
--- a/langkit/support/langkit_support-lexical_env.adb
+++ b/langkit/support/langkit_support-lexical_env.adb
@@ -1999,17 +1999,7 @@ package body Langkit_Support.Lexical_Env is
    function Is_Stale (Self : Lexical_Env) return Boolean is
       L : Lexical_Env;
    begin
-      if Self.Owner /= No_Unit then
-         --  If there is an owner, check that the unit version has not been
-         --  incremented since then. Additionally, if this is a Rebound env,
-         --  check that the version of the context has not been incremented
-         --  either, to make sure that the rebindings stored in this env
-         --  are still valid.
-         return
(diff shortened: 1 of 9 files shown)


langkit/support/langkit_support-lexical_env.adb
```diff
@@ -1998,7 +2000,10 @@ package body Langkit_Support.Lexical_Env is
       L : Lexical_Env;
    begin
       if Self.Owner /= No_Unit then
-         return Get_Version (Self.Owner) > Self.Version;
+         return
+            Get_Version (Self.Owner) > Self.Version or else
```
raph-amiard: So the logic here is: 

If there is an owner for the env:

1. In any case, the unit version of the lexical env owner needs to not be stale
2. If thee env is a rebound env, the context version needs to not be stale

I suggest that:

* You comment along those lines because the code is not very readable as-is. (including the part about there being an owner, even if you didn't write it).
* You rename `Get_Version` and `Version` into `Get_Unit_Version` and `Unit_Version` respectively, since now there are several versions, it will make the code clearer.

EDIT: I cannot comment on code you didn't write, but commenting the `when Rebound` branch below to say "if there is no owner, we only care about the fact that the rebound_env is not stale".

Also, I believe your code is "accidentally" correct here: You don't check if the rebound env is stale, but this will never happen because it cannot be stale if the context_version isn't. This might be worth commenting also!

PR #213, opened 2019-09-03
Decouple public Ada API and implementation + testsuite perf enhancements

pmderodat:


langkit/emitter.py
```diff
@@ -356,6 +357,9 @@ def __init__(self, template_base_name, rel_qual_name,
                 :param bool ada_api: Whether we can avoid generating this unit
                     if the Ada API is disabled.
 
+                :param bool unparser: Whether we can avoid generating this unit
```
raph-amiard: While you did the work for the testsuite, you should make those options (generate_unparser and generate_ada_api) available to the users, via config flags/command line options, since they're of global interest for langkit users.

pmderodat: Agreed. I did that: see libmanage.py:197 for `--no-ada-api` and we already do not generate unparsers unless explicitly requested (`--generate-unparser`).

raph-amiard: Ah cool :) My bad, missed that

PR #214, opened 2019-09-05
Tentative introduction of stack overflow mitigation

pmderodat:


pmderodat:
I’ll answer your three points separately.

# Parsers
IIRC, we have this requirement that LAL must not crash, even if you run it on binary files. Stack overflow can happen easily in parsers if for some reason a source file contains too many recursive constructs, such as a user that accidentally keeps inserting `(` because of a keyboard issue. If we agree at least on that, I can measure performance with mitigation only generated there. I don’t think maintainance will be a problem there as the instrumentation itself in parsers is restricted to a small number of changes in `parsers/fn_code_ada.mako`, which is a portion of the code that seldom evolves.

# Populate lexical env (PLE)
I’m assuming that we agreed in the previous section: we consider that LAL must be resilient to deep parse trees, and in particular here, deep enough to start causing troubles but not deep enough to trigger a stack overflow during parsing.

PLE is just a parse tree traversal calling several properties. With deep trees, we can easily imagine tree traversal consuming more than half of the stack budget and properties eating the rest, triggering a stack overflow. Considering that here as well, instrumentation is not very invasive and will not create maintainance burden (same kind of changes as in parsers: localized and targetting stable portions of templates), I see no good reason to remove the protection there if we see that it’s not where we loose most of performance. Just like parsers, I could also check performance with mitigation only generated there.

# Lexical environment lookups
First, I would like to mention that when I started this work, I did not expect to modify this area that much. Recursivity is very widespread in `Langkit_Support.Lexical_Envs`: lookups, but also `Env_Getter` resolution, which is used everywhere. That being said:

* We don’t need cyclic env structures nor infinite recursions inside properties to trigger stack overflows during the lexical environment lookups. Actually my first attempt at writing a testcase for a stack overflow that involved both properties and lexical env lookups yielded an infinite recursion in `Env_Getter` resolution: see the `recursive_lookup` part of the testcase I added. I can imagine even more ways to do it that don’t involve recursion in properties. If I understood correctly our live discussion yesterday, you did see such an infinite recursion yourself in the past while working in Libadalang. So we have evidence that stack overflow protection here could be useful in practice.

* Even though instrumentation is invasive, it’s easy to grasp, and thus (I think) far easier to maintain than the rest of Lexical_Envs’s complex logic. The principle is very simple: call `Enter_Call` at the beginning of a re-entrant function, then call `Exit_Call` on all exit points (which are obvious to find syntactically: exceptions handlers and `return` statements. [1] Note: for now, my position regarding the this module is that it’s already more complex that I’m comfortable with, but it’s complexity is there so that we can make name resolution work for Ada and reduce complexity in the DSL (it’s a tradeoff). Right now, I consider that the extra (small IMO, but of course I’m biased) complexity for stack overflow protection is worth it.

[1] You said:
> This protection will only be as good as people maintaining it: it's not automatic (and thus, useless in this part of the code IMHO).

I disagree. First, it’s not because a protection *can* be accidentally disabled by further contributions that it’s useless. There is an easy way to avoid this: testing (and my testcase is a first step in this direction). Actually if this is the main blocking point, I would be even willing to write a LAL-based tool to check this, and the maintainance problem would be mostly gone! Also, I realize now that I should probably add a section in Langkit’s dev manual to describe the principle of the protection to future contributors. I’ll do that if we end up agreeing that this protection is worth it.

raph-amiard:
For me there is a simple heuristic here: Which part of your work mitigates the original stack overflow ?

The answer is: The one in properties. All the other use cases you presented are theoretical - by that I mean not extracted from bugs reported by users - and thus, not things I'm comfortable pushing *at the same time*, and in urgency just before the release, regardless of their technical merits.

pmderodat:
> For me there is a simple heuristic here: Which part of your work mitigates the original stack overflow ?

As discussed live: we’re very close to the release, so we’ll keep the protection as minimal as possible (i.e. only for properties) and pay the associated performance cost, considered to be worth it. This is motivated by the fact that, beyond the bug that started this work, all stack overflows we have seen in practice so far involved recursions involving properties; in other words: we are not aware of any bug report for stack overflows that this restricted protection would not have caught.

I’ll update this PR to be merged before the release and put the original work in a branch (https://github.com/pmderodat/langkit/tree/stack-overflow). We’ll get back to this more general protection after the release.

langkit/templates/pkg_implementation_spec_ada.mako
```diff
@@ -52,6 +52,26 @@ private package ${ada_lib_name}.Implementation is
    type Analysis_Context_Type;
    type Internal_Context is access all Analysis_Context_Type;
 
+   Unexpected_Call_Depth : exception;
```
raph-amiard: This is not used

pmderodat: Nice catch: I’ve pushed an update to use it. Thanks!

PR #218, opened 2019-09-16
S819-034: Fix determinism issue of env get.

Roldak:
In the special scenario where "get" is called with the "null" symbol,
the output was not deterministic. This commit makes sure it is.

### DIFF
--- a/langkit/support/langkit_support-lexical_env.adb
+++ b/langkit/support/langkit_support-lexical_env.adb
@@ -21,6 +21,7 @@
 -- <http://www.gnu.org/licenses/>.                                          --
 ------------------------------------------------------------------------------
 
+with Ada.Containers.Generic_Array_Sort;
 with Ada.Strings.Unbounded;           use Ada.Strings.Unbounded;
 with Ada.Strings.Wide_Wide_Unbounded; use Ada.Strings.Wide_Wide_Unbounded;
 with Ada.Text_IO;                     use Ada.Text_IO;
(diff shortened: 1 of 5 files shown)


testsuite/tests/lexical_envs/env_get_all/test.py
```diff
@@ -0,0 +1,43 @@
+"""
+Test add_to_env with multiple env_assoc having different destination
+environments.
```
pmderodat: Please update the docstring

testsuite/tests/lexical_envs/env_get_all/test.py
```diff
@@ -0,0 +1,43 @@
...
+"""
+
+from __future__ import absolute_import, division, print_function
+
+from langkit.dsl import (ASTNode, T, has_abstract_list)
```
pmderodat: Unrequired parens

langkit/support/langkit_support-lexical_env.adb
```diff
@@ -574,11 +575,59 @@ package body Langkit_Support.Lexical_Env is
...
+      ----------------------
+
+      procedure Append_All_Nodes
+        (Env : Lexical_Env; From_Rebound : Boolean)
+      is
+         type Cursor_Array is array
+           (Natural range <>) of Internal_Envs.Cursor;
+
+         function "<" (A, B : Internal_Envs.Cursor) return Boolean is
+           (Internal_Envs.Key (A).all < Internal_Envs.Key (B).all);
```
pmderodat: There is still nondeterminism when we have multiple entries with the same keys, right? At this point I wonder: maybe we should define a total order for nodes just for internal use (it has no special meaning: we just want deterministic order) that would order on 1) the unit filename 2) the node sloc_range. What do you think?

testsuite/tests/lexical_envs/env_get_all/main.py
```diff
@@ -0,0 +1,38 @@
...
+
+def test_case(name, content):
+    u = ctx.get_from_buffer(name, content)
+    if u.diagnostics:
+        for d in u.diagnostics:
+            print(d)
+        sys.exit(1)
+
+    u.populate_lexical_env()
+    print(" ".join(str(x.text) for x in u.root.p_env_get_all))
```
pmderodat: Let’s use single quotes (for consistency)

langkit/support/langkit_support-lexical_env.adb
Roldak: Mhh I'm not sure what you mean. The indeterminism I am trying to fix here is the indeterminism that is inherent to the implementation of a hashtable in GNAT. But those are not "multi-hashmaps" so "multiple entries with the same keys" is not possible. I guess you are wondering what happens when we add several nodes with the same key (say, K) in the same env (say, E). Well, in that case the order is well defined, and it's the order in which populate lexical env operates (which AFAIK is deterministic), which will append the vector associated to key K in env E with the new node each time the add_to_env action is executed.

Roldak: However you made me realize that it would be a good idea to test the addition of multiple nodes with the same key, so I will update my testcase accordingly :)

raph-amiard: FWIW my understanding is the same as Romain's, ie. there is no indeterminism when you have multiple entries with the same key: If they're in the same env, they're order (by insertion order). If they're in different envs, they're still ordered (by traversal order).

PR #224, opened 2019-09-24
S924-008 Fix unwrapping of enum types

danielmercier:
An EnumType has a ctype view, thus, to unwrap it, it is unecessary to
call the unwrap function.

The wrap function was not being affected by the issue.

wrap and unwrap functions are required only if a type does not define a
ctype view. Thus, refactor the code to reflect this.

### DIFF
--- a/langkit/ocaml_api.py
+++ b/langkit/ocaml_api.py
@@ -253,6 +253,34 @@ class OCamlAPISettings(AbstractAPISettings):
             ' (wrapping): {}'.format(type)
         ))
 
+    def has_ctype_view(self, type):
+        """
+        Given a type, return true if it has a ctype view. Having a ctype view
+        means that wrap and unwrap operations are not needed explicitely.
+
+        :param ct.CompiledType type: The type to check.
+        :rtype: bool
+        """
+
+        return dispatch_on_type(type, [
+            (T.AnalysisUnit, lambda _: False),
+            (ct.EnumType, lambda _: True),
+            (ct.ASTNodeType, lambda _: True),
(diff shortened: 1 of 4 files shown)


langkit/ocaml_api.py
```diff
@@ -253,6 +253,35 @@ def wrap_requires_context(self, type):
             ' (wrapping): {}'.format(type)
         ))
 
+    def have_ctype_view(self, type):
```
pmderodat: I suggest to rename to `has_ctype_view`: we check for a single type, so singular would make more sense

langkit/ocaml_api.py
```diff
@@ -253,6 +253,35 @@ def wrap_requires_context(self, type):
...
+        """
+        Given a type, return true if it has a ctype view. Having a ctype view
+        means that wrap and unwrap operations are not needed explicitely.
+
+        :param ct.CompiledType type: Type we want to check for having a view
+            or not.
```
pmderodat: “The type to check.”?

pmderodat:
Hello Daniel,

Thank you very much for this fix! Do you think you could enhance the `ocaml_api/general` testcase to cover the case your patch fixes?

langkit/ocaml_api.py
danielmercier: Sure.. Done

langkit/ocaml_api.py
danielmercier: Maybe my sentence was a bit to complex, agreed :p

danielmercier:
Thank for the review, I made the changes and added the test to ocaml_api/general

PR #226, opened 2019-10-08
Fix memory leaks in OCaml bindings

danielmercier:


langkit/ocaml_api.py
```diff
@@ -369,6 +369,29 @@ def is_struct(self, type, from_module=None):
             (T.EnvRebindings, lambda _: False),
         ])
 
+    def get_finalize(self, type):
+        """
+        Return the name of the finalization function if a value of the given
+        type needs a finalization. A value needs a finalization if it is
+        completly converted to an OCaml function. Thus unwrapping
```
pmderodat: Sorry for this newbie question :) Are the values for `type` really converted to functions? (that sounds weird: how does wrapping/unwrapping work on functions?)

langkit/ocaml_api.py
```diff
@@ -369,6 +369,29 @@ def is_struct(self, type, from_module=None):
...
+        the value, it is required to finalize it after using it.
```
pmderodat: That part of the comment is unclear to me. The logic behind wrapping/unwrapping is that one one hand, you have types for C API consumption (internal types in the OCaml bindings), and on the other hand you have public OCaml types for OCaml user consumption. Wrapping is the action of creating a public OCaml value out of C API value, and of course unwrapping is the opposite.

If OCaml bindings work like Python’s bindings, finalization happens in two cases:

1. We allocated a C API value when unwrapping a public value to pass it as a C API function parameter: in this case we need to finalize this C API value after the function returns.
2. A public value is deallocated by Python’s runtime, so we need to finalize the C API value that it wraps.

Which case is covered here? Both? If I understand the comment correctly, this targets only 1.

langkit/templates/ocaml_api/struct_types_ocaml.mako
```diff
@@ -256,35 +249,42 @@ module AnalysisContextStruct : sig
 end = struct
   (* Module defining the c structure of an analysis context *)
 
-  type t = unit ptr
-  let c_type = ptr void
+  (* The real C type of a context is a void*. But we use a pointer to this
+     type, to be able to allocate a value of t and attach a finaliser to it.
```
pmderodat: finalizer* (there are other occurences)

testsuite/python_support/utils.py
```diff
@@ -264,9 +264,13 @@ def run(*argv):
         with open('dune-project', 'w') as f:
```
pmderodat: The commit content and its message disagree with each other. :-) I suggest to split this commit in three:

* add a Valgrind suppression file for OCaml programs
* python_support/utils.py: build and run OCaml programs separately
* ocaml_api/general: enhance to test BigInt bindings

langkit/ocaml_api.py
danielmercier: Not a newbie question at all, that was a typo. We create OCaml values and not functions

langkit/ocaml_api.py
danielmercier: I modified the comment to clarify that we are talking about (1.) indeed.

danielmercier: I added some types that are not fully converted to OCaml values (they still keep the C value)

langkit/ocaml_api.py
pmderodat: Ok, thanks!

langkit/ocaml_api.py
pmderodat: Understood, thanks!

testsuite/python_support/utils.py
pmderodat: Thanks a lot for the extra explanations in the new commits, they allowed me to understand changes much better. :-)

PR #229, opened 2019-10-24
Make arrays hashable.

Roldak:


pmderodat:
Thanks for the experiment! This is superseeded by #231

PR #230, opened 2019-10-24
Add dump_ir option on langkit properties

raph-amiard:


langkit/expressions/base.py
```diff
@@ -3308,6 +3308,11 @@ def __init__(self, expr, prefix, name=None, doc=None, public=None,
 
         self.activate_tracing = activate_tracing
 
+        self.dump_ir = dump_ir
+        """
+        If true, dump the tree of resolved expressions for this property.
```
pmderodat: Can you please move this to the function docstrings? (where we document all params)

raph-amiard: Sure

PR #232, opened 2019-10-30
SA30-024 ocaml bindings: make entity types optional in struct types

danielmercier:


langkit/ocaml_api.py
```diff
@@ -305,7 +306,7 @@ def from_module(typ, value):
         else:
             return wrapped_result
 
-    def unwrap_value(self, value, type, context):
+    def unwrap_value(self, value, type, context, check_for_none=False):
```
pmderodat: Please add a `check_for_none` entry to the docstring below

langkit/ocaml_api.py
```diff
@@ -328,9 +329,17 @@ def from_module(typ, value):
...
             )
 
-        if isinstance(type, ct.EntityType):
+        if type.is_entity_type:
             # For entity type, the unwrap function is inside the root node
-            return from_module(T.root_node, value)
+            if check_for_none:
+                fmt_check = (
+                    "match {} with Some n -> {} "
+                    + "| None -> make EntityStruct.c_type"
```
pmderodat: Python allows you to have consecutive string literals: they are considered one big literal. So you could write instead:

```python
fmt_check = ('match {} with'
             ' Some n -> {}'
             ' | None -> make EntityStruct.c_type')
```
and given that the return statement is right after:

```python
return ('match {} with'
        ' Some n -> {}'
        ' | None -> make EntityStruct.c_type'
        .format(value, from_module(T.root_node, 'n'))
```

langkit/templates/ocaml_api/struct_types_ocaml.mako
```diff
@@ -23,7 +23,15 @@ end
...
       % else:
+         % if f.type.is_entity_type:
+            ## If the type is an entity type, we want an optional type instead.
+            ## This is because we have no way to know if the node will be null
+            ## or not.
+    ${ocaml_api.field_name(f)} :
+      ${ocaml_api.type_public_name(f.type, cls)} option;
+         % else:
     ${ocaml_api.field_name(f)} : ${ocaml_api.type_public_name(f.type, cls)};
+         % endif
```
pmderodat: There is a lot of code duplication, here: may I suggest:

```ocaml
${ocaml_api.field_name(f)} :
  % if f.type.is_ast_node:
    ${ocaml_api.c_value_type(f.type, cls)}
  % elif f.type.is_entity_type:
    ## If the type is an entity type, we want an optional type instead.
    ## This is because we have no way to know if the node will be null
    ## or not.
    ${ocaml_api.type_public_name(f.type, cls)} option
  % else:
    ${ocaml_api.type_public_name(f.type, cls)};
  % endif
;
```

danielmercier:
> These changes take care of wrapping entity types in option types so that we can represent separately entities that can be null from entities that can’t. However I don’t understand how this should work for bare nodes: we generally wrap them into entities in public APIs. It seems that we don’t do this in the OCaml bindings: can you shade some light on bare nodes handling?

In OCaml bindings, bare nodes are not converted into any OCaml value. Instead, they are kept as pointers. Thus, a null bare node is a null pointer in OCaml.

langkit/templates/ocaml_api/struct_types_ocaml.mako
pmderodat: Given that OCaml is not a indentation-sensitive language, I think it’s more readable for templates to be written so that indentation for the Mako syntax prevails over indentation for the OCaml code: here it’s hard to understand the structure of the code looking at the indentation, whereas conceptually there are if/elif/else controls in which all clauses emit a type expression. What do you think?

pmderodat:
Thanks for the rationale! Fair enough, it’s not the goal of this PR to change formatting policy. Using ocamlformat sounds interesting (this is what we do for generated Ada and Python sources), but let’s do that later.

PR #237, opened 2019-11-13
SB13-013: Allow predicates on the root AST node.

Roldak:


testsuite/tests/properties/top_level_predicate/test.py
```diff
@@ -0,0 +1,35 @@
+"""
+Test that Prediate works well with default argument values.
```
pmderodat: Please update the docstring.

PR #249, opened 2019-12-16
Memoization fix, source buffer decoding improvements

pmderodat:


langkit/templates/memoization_ada.mako
```diff
@@ -87,8 +87,7 @@ function Find_Memoized_Value
 procedure Add_Memoized_Value
```
raph-amiard: Typo in the commit message "exceution"

raph-amiard: Also I think this would warrant a change entry in the LAL repo since this had potential repercussions on LAL users

langkit/documentation.py
```diff
@@ -374,17 +371,19 @@ def instantiate_templates(doc_dict):
         Return a hash for this unit, to be used in hash tables.
     """,
     'langkit.unit_reparse_file': """
```
raph-amiard: I'm actually wondering if sticky behavior in this one case wouldn't be good to keep: when you parsed from a file on disk before. Because, you're not directly passing a buffer, which was the original thing I warned about being dangerous in the context of the Python API, and in that case forgetting the old encoding seems weird and a bit counter intuitive.

No strong feeling either way though

pmderodat:
Split in two PRs: #250 and #251

PR #250, opened 2019-12-18
Source buffer decoding improvements

pmderodat:


langkit/documentation.py
```diff
@@ -374,17 +371,19 @@ def instantiate_templates(doc_dict):
         Return a hash for this unit, to be used in hash tables.
     """,
     'langkit.unit_reparse_file': """
```
pmderodat: Continuing the discussion on https://github.com/AdaCore/langkit/pull/249#discussion_r359353498
> I'm actually wondering if sticky behavior in this one case wouldn't be good to keep: when you parsed from a file on disk before. Because, you're not directly passing a buffer, which was the original thing I warned about being dangerous in the context of the Python API, and in that case forgetting the old encoding seems weird and a bit counter intuitive.

I’m sure sure I understand what you suggest. Let’s talk about each case separately:
* what charset should we use when calling `get_from_buffer` without providing one? The context-wide default? The previous charset passed to `get_from_file`? Or the default one passed to `get_from_buffer`?
* Same question for `get_from_file`.

If that doesn’t matter much from a useability point of view, I would advocate to keep rules simple: no buffer passed => use context-wide charset. This is probably simpler to reason about for both us and users.

PR #256, opened 2019-12-19
$.Analysis: enhance the formatting of generated property declarations

pmderodat:


pmderodat:
For instance: on Libadalang, we used to have:
```ada
   function P_Index_Type 
  (Node : Base_Type_Decl'Class
      ; Dim : Integer

      

      ; Origin : Ada_Node

      'Class

         := No_Ada_Node
  ) return Base_Type_Decl
;
   --  Return the index type for dimension ``dim`` for this type, if
   --  applicable.
```
while we now have instead:
```ada
   function P_Index_Type
     (Node : Base_Type_Decl'Class;
      Dim : Integer;
      Origin : Ada_Node'Class := No_Ada_Node) return Base_Type_Decl;
   --  Return the index type for dimension ``dim`` for this type, if
   --  applicable.
```

PR #257, opened 2020-01-02
Make Lexical_Env.Get return a vector

raph-amiard:


langkit/support/langkit_support-vectors.adb
```diff
@@ -106,6 +106,22 @@ package body Langkit_Support.Vectors is
...
+   -- Concat --
+   ------------
+
+   procedure Concat (Self : in out Vector; Elements : Vector) is
+   begin
+      if Small_Vector_Capacity /= 0 then
+         Self.Concat (Elements.To_Array);
+      else
+         for I in Elements.First_Index .. Elements.Last_Index loop
+            Self.Reserve (Self.Length + Elements.Length);
```
Roldak: why is this in the loop?

raph-amiard: Because mistake. Good catch!

PR #259, opened 2020-01-02
Fix CE when calling prop via introspection 

raph-amiard:


raph-amiard:
@pmderodat thanks, done in the libadalang PR https://github.com/AdaCore/libadalang/pull/412

PR #260, opened 2020-01-02
T102-023: don't automatically log properties on STDOUT

raph-amiard:
.. and make sure that users of C/Python/OCaml APIs have a way to
configure traces, by parsing the traces config file when using those
APIs.


langkit/templates/pkg_init_body_ada.mako
```diff
@@ -19,6 +20,14 @@ package body ${ada_lib_name}.Init is
...
          raise Program_Error with "Libiconv is not available";
       end if;
+
+      --  We'll also initialize GNATCOLL.Traces here. It seems preferable to do
+      --  this here rather than in elaboration: Ada API users won't be forced
+      --  to use the config file, if they're GNATCOLL.Traces users with a
+      --  different workflow. However, everybody else will (Python, C, OCaml
+      --  API users), which seems like a good compromise.
+
+      GNATCOLL.Traces.Parse_Config_File;
```
pmderodat: Actually, the fact that our current Ada programs don’t call this procedure is an oversight. I have a commit locally to fix that, making the call to `Initialize` on elaboration. The documentation of this public API already says that it must be called before anything else in LAL/Langkit_Support, so preventing multiple calls to Initialize looks like a breaking API change. What about creating a new API entry point instead?

pmderodat: Another reason we would want to create a dedicated entry point in bindings for this is that this will play badly with Ada programs embedding a Python interpreter that loads LAL (for instance GS): the call to Parse_Config_File will conflict with GS’.

raph-amiard: > Actually, the fact that our current Ada programs don’t call this procedure is an oversight. I have a commit locally to fix that, making the call to Initialize on elaboration. 

I don't understand what you mean. It's definitely not an oversight that it is not called during the elaboration of Libadalang, I considered it and I don't like this solution. `GNATCOLL.Traces` might be used in client applications that don't use config files, and that might create confusing situations.

> The documentation of this public API already says that it must be called before anything else in LAL/Langkit_Support, so preventing multiple calls to Initialize looks like a breaking API change. What about creating a new API entry point instead?

I don't understand this comment either. The call to `Parse_Config_File` can be done multiple times without any problems, what would make you think otherwise?

raph-amiard: > Another reason we would want to create a dedicated entry point in bindings for this is that this will play badly with Ada programs embedding a Python interpreter that loads LAL (for instance GS): the call to Parse_Config_File will conflict with GS’.

Hmm, indeed, it might be called with different parameters in GNATStudio. However, adding a new entry point in bindings is way more work than I'm ready to put into this given that it's only of theoretical importance at the moment

raph-amiard:
As discussed live, I removed the code in `Initialize`. @pmderodat will get rid of it altogether, and at a later stage we'll add an API entry point to init traces from other APIs than Ada.

testsuite/tests/properties/logging/test.py
```diff
@@ -62,5 +62,5 @@ class Ref(FooNode):
     ref=Ref(fg.name),
     name=Name(Token.Identifier),
 )
-build_and_run(fg, 'main.py')
+build_and_run(fg, ada_main='main.adb')
```
pmderodat: Could you also remove `main.py`? Thanks!

raph-amiard: Argh sorry, yes

pmderodat: Thanks! :-)

PR #262, opened 2020-01-03
Add basic code coverage report capabilities

pmderodat:


langkit/templates/pkg_init_spec_ada.mako
```diff
@@ -2,13 +2,9 @@
...
-
-   procedure Initialize
-     with Export        => True,
-          Convention    => C,
-          External_Name =>
-             "${ada_lib_name.lower()}_initialize";
-   ${ada_doc('langkit.initialize', 3)}
+private package ${ada_lib_name}.Init is
+
+   procedure Initialize;
```
raph-amiard: I'm wondering, is there really a need to have a separate package (with a fake procedure in the spec) just for a check for Iconv? Why not just inline that in `Implementation` or `Analysis` or `Common` or whatever else needs a body ?

langkit/templates/project_file.mako
```diff
@@ -22,21 +34,75 @@ library project ${lib_name} is
...
+   ## Disable style checks on instrumented code
+   % if emitter.coverage:
+      Enable_Warnings : Boolean := "false";
+   % else:
+      Enable_Warnings : Boolean :=
+        external ("${lib_name.upper()}_WARNINGS", "false");
+   % endif
+
+   ## We rely a lot on project installation (gprinstall), and gnatcov does not
+   ## handle well that, so we need kludges to compute code coverage using
```
raph-amiard: "does not handle well that" -> "does not handle that well"

langkit/templates/pkg_init_body_ada.mako
```diff
@@ -2,10 +2,53 @@
...
+
+with GNATcov_RTS.Buffers.Lists.${ada_lib_name};
+with GNATcov_RTS.Traces.Output;
+% endif
+
 package body ${ada_lib_name}.Init is
 
    procedure Initialize is null;
 
+   % if emitter.coverage:
```
raph-amiard: OK you can probably disregard my earlier comment about `Init` then ^^

langkit/gdb/debug_info.py
```diff
@@ -8,7 +8,12 @@
 import inspect
```
raph-amiard: That's a cosmetic comment, but the naming scheme becomes mildly unclear IMHO. We can maybe see that when we have to rework this for the concrete syntax, but the part that are basically "source maps" go beyond gdb (that's the point of this commit), so having those in a module like `dbg_info.source_maps` rather than `gdb.debug_info` would make more sense IMHO (and then keep the purely gdb part in the `gdb` module)

langkit/templates/pkg_init_spec_ada.mako
pmderodat: We could move that to Implementation or Common, indeed (Analysis may not be generated nowadays). I’ll do that in another commit, thanks.

langkit/templates/pkg_init_body_ada.mako
pmderodat: We still can get rid of this package. ;-) I’ll just do it as another step to avoid messing with the commit history.

langkit/gdb/debug_info.py
pmderodat: I agree it woud make more sense. I’ll do that in other commits.

PR #263, opened 2020-01-06
Fix a memory leak

raph-amiard:


langkit/support/langkit_support-vectors.adb
```diff
@@ -50,22 +50,34 @@ package body Langkit_Support.Vectors is
...
-            Self.E := To_Pointer (Alloc (Siz));
-            for I in Self.SV'Range loop
-               Self.E.all (I) := Self.SV (I);
-            end loop;
-         else
-            Self.E := To_Pointer (Realloc (Self.E.all'Address, Siz));
+      if Small_Vector_Capacity > 0
+        and then Self.Capacity = Small_Vector_Capacity
+      then
+         --  We have an inline small vector, and we're still using it.
```
pmderodat: Stylecheck issue

langkit/support/langkit_support-vectors.adb
```diff
@@ -50,22 +50,34 @@ package body Langkit_Support.Vectors is
...
-      if Small_Vector_Capacity > 0 then
-         if Self.Capacity = Small_Vector_Capacity then
-            Self.E := To_Pointer (Alloc (Siz));
-            for I in Self.SV'Range loop
-               Self.E.all (I) := Self.SV (I);
-            end loop;
-         else
-            Self.E := To_Pointer (Realloc (Self.E.all'Address, Siz));
+      if Small_Vector_Capacity > 0
+        and then Self.Capacity = Small_Vector_Capacity
```
pmderodat: Just in case: have you checked the performance impact of this? IIRC, we had two separate IF blocks because when the vector is instanciated with `Small_Vector_Capacity => 0`, GCC would fold the always false `Small_Vector_Capacity > 0` condition only if it was a standalone condition in an IF block.

raph-amiard: I didn't but I'm really dubious that this is why the code was written that way, I think it was written that way for correctness reasons in that particular case (the `Self.Capacity := Capacity` being at the end)

raph-amiard: Just checked: no performance difference

PR #275, opened 2020-01-10
Work on unparsing comments

Roldak:


langkit/dsl_unparse.py
```diff
@@ -1,10 +1,177 @@
...
+            return self
+
+        def __exit__(self, exc_type, exc_val, exc_tb):
+            return self
+
+        def emit_comments(self):
+            return ""
+
+    @staticmethod
+    def class_from_location(loc):
```
raph-amiard: Plz add a docstring

langkit/dsl_unparse.py
```diff
@@ -1,10 +1,177 @@
...
+    @contextmanager
+    def _with_current_node(self, new_node):
+        old_node = self.current_node
+        self.current_node = new_node
+        self.current_token = new_node.token_start
+        yield
+        self.current_node = old_node
+        self.current_token = new_node.token_end
+
+    def property(self, prop):
```
raph-amiard: Plz add a docstring for non annotated context managers saying that they're context managers

langkit/dsl_unparse.py
```diff
@@ -118,25 +285,45 @@ def is_simple_expr(expr):
...
     from langkit.expressions import Let
-    if is_simple_expr(expr):
-        strn = emit_expr(expr, **ctx)
-        return emit_paren(strn) if len(strn) > 40 else strn
-    elif isinstance(expr, Let):
+
+    if isinstance(expr, Let):
         return emit_expr(expr, **ctx)
+
+    strn, has_coms = prepend_comments(expr, **ctx)
```
raph-amiard: Not bloçcking but it would be good to understand why the call to `prepend_comments` is made inside `emit_paren_expr`

langkit/utils/__init__.py
```diff
@@ -174,6 +175,23 @@ def copy_to_dir(filename, dirname):
     shutil.copy(filename, os.path.join(dirname, os.path.basename(filename)))
 
 
+def add_search_path(name, path):
+    """
+    Return as a string the Bourne shell commands that add the given path
```
pmderodat: that adds*

make.py
```diff
@@ -0,0 +1,252 @@
...
+from __future__ import absolute_import, division, print_function
+
+from contextlib import contextmanager
+
+import argparse
+
+import os
+import shutil
+import subprocess
+import sys
```
pmderodat: Merge this import block with the two previous ones?

make.py
```diff
@@ -0,0 +1,252 @@
...
+    copy_all_shared_libs(prefix, dest, lang_name, "langkit_support")
+
+    pip_install(lang_python_dir)
+
+
+def build(args):
+    """
+    Build langkit and its dependencies.
+
+    :type argparse.Namespace
```
pmderodat: Misformatted type annotation

make.py
```diff
@@ -0,0 +1,252 @@
...
+    yield
+    os.chdir(old_dir)
+
+
+def pip_install(setup_dir):
+    """
+    Install the python package located in the given directory.
+
+    :type setup_dir: str
+    :raise CalledProcessError if installation terminates with non-zero status
```
pmderodat: I don’t think this is a correct Sphinx annotation. May I suggest moving this to a regular English sentence after the description? This is the style we have been using in Langkit to describe exceptions.

make.py
```diff
@@ -0,0 +1,252 @@
...
+    Copy all shared libraries from anywhere in a given source tree to a given
+    destination folder.
+
+    Only the libraries that satisfy the given simple patterns are copied.
+    Given patterns must NOT be regular expressions, but simple strings.
+    A library will match if the pattern is a substring of the library's name.
+
+    :type from_dir: str
+    :type destination: str
+    :type patterns: *str
```
pmderodat: I think this should be: `type patterns: list[str]`

make.py
```diff
@@ -0,0 +1,252 @@
...
+                    shutil.copy(os.path.join(root, filename), destination)
+
+
+def manage(lang_dir, verb, build_mode, *add_args, **kwargs):
+    """
+    Run the manage.py script from of a Langkit-based language definition.
+
+    :param str verb: 'generate', 'build', 'install', etc.
+    :param str build_mode: 'dev' or 'prod'.
+    :param str|None rpath: if set, defines the RPATH to include in built
```
pmderodat: If `rpath` is documented, why not moving it out of `kwargs`? And why have `kwargs` at all?

Roldak: I wanted `rpath` to be an optional argument, but also allow additional arguments to pass to `manage.py` through `*args`. I'm not sure how to achieve that any other way... Or do you prefer I remove the `*args` and ask for a list explicitly?

pmderodat: I see… thanks. Actually given that we pass `-R` nightly, I would suggest to remove this altoghether and always build with `--gargs=-R`.

Roldak: Can it not be useful for a development build though to keep the rpaths? (For example if you want to `build` and `setenv` to develop on langkit)

pmderodat: Not really: for the Python bindings, running the install part, which bundles all libs thus removing the need to fiddle with RPATHs, is very cheap, and for Ada programs, these can be built with RPATHs (happens by default), so the RPATHs in libraries are useless. So I really think we could remove it here :-)

make.py
```diff
@@ -0,0 +1,249 @@
...
+setenv_parser = subparsers.add_parser('setenv')
+setenv_parser.set_defaults(command='setenv')
+
+install_parser = subparsers.add_parser('install')
+install_parser.set_defaults(command='install')
+install_parser.add_argument('prefix', type=str,
+                            help='Where to install Langkit dependencies')
+install_parser.add_argument('--gnat-install-dir', type=str, default=None,
+                            help='Specify where GNAT is installed. If not'
+                                 'provided, assume it is installed where '
```
pmderodat: A space is missing between "not" and "provided". We never decided it was as a rule, yet I advise to put the spaces at the beginning of the next line to make it clear when it’s missing.

make.py
```diff
@@ -0,0 +1,249 @@
+"""
+Helper script to build langkit and its dependencies.
+"""
+
+from __future__ import absolute_import, division, print_function
+
+from contextlib import contextmanager
```
pmderodat: The fact that the stylechecker does not complain about this line makes me realize that the stylechecker does not run on this script! Please move it to the `scripts` directory.

make.py
```diff
@@ -0,0 +1,249 @@
...
+    :param str build_mode: 'dev' or 'prod'.
+    :param list[str] add_args: additional arguments to pass
+    """
+    with in_dir(lang_dir):
+        args = [sys.executable, 'manage.py', verb]
+        args.extend(['--build-mode', build_mode])
+        args.extend(add_args)
+
+        if verb in ('build', 'make'):
+            # Always build shared libs with "$ORIGIN" rpaths.
```
pmderodat: IIUC, `-R` disables run paths, so this comment is wrong. I suggest instead:
> Do not use run path options
(from `gprbuild --help`)

make.py
```diff
@@ -0,0 +1,249 @@
...
+            args.append('--gargs=-R')
+            env = os.environ.copy()
+            env['LD_RUN_PATH'] = "$ORIGIN"
```
pmderodat: Since we don’t use run path, you can remove all env modifications, here.

make.py
```diff
@@ -0,0 +1,249 @@
...
+    pip_install(lang_python_dir)
+
+
+def build(args):
+    """
+    Build langkit and its dependencies.
+
+    :type args: argparse.Namespace
+    """
+    # build Libpythonlang
```
pmderodat: Missing capital (Build); same several times below

utils/travis-install.sh
```diff
@@ -92,14 +92,9 @@ gcc -v
...
+python make.py install "$LIB_INSTALL_DIR" --gnat-install-dir="$INSTALL_DIR"
 
 # Also install flake8 to have complete style checks
 pip install flake8
 
-# RA22-015: Install libpythonlang to make it available to dsl_unparse
-(
-    cd ./contrib/python
-    ./manage.py make -P
-)
```
pmderodat: Now that libpythonlang is built and installed, could you update `tests/contrib/python_parse_1` not to recompile it? (just run `parse` from the environment)

make.py
```diff
@@ -0,0 +1,249 @@
...
+
+parser = argparse.ArgumentParser(
+    description='Helper script to build langkit and its dependencies'
+)
+parser.add_argument('--build-mode', default='dev')
+
+subparsers = parser.add_subparsers()
+
+build_parser = subparsers.add_parser('build')
+build_parser.set_defaults(command='build')
```
pmderodat: I suggest you move the parsers declaration after the function definitions and add `func=XXX` (`func=build`, `install` and `setenv`) to all subparsers? This will make manual dispatching starting line 242 unecessary. See similar processing in `libmanage.py`.

make.py
```diff
@@ -0,0 +1,249 @@
...
+    manage(lang_dir, 'install', build_mode, prefix)
+
+    # Copy all necessary shared libraries in the python package, so that the
+    # subsequent call to pip_install packages all of them.
+
+    lang_python_dir = os.path.join(lang_dir, 'build', 'python')
+    dest = os.path.join(lang_python_dir, lang_name)
+
+    copy_all_shared_libs(gnat_install, dest, "gnat", "gnarl", "xmlada")
+    copy_all_shared_libs(gnatcoll_bindings_install, dest, "gnatcoll")
```
pmderodat: This is not complete: both xmlada and gnatcoll-core can be installed outside of gnat. I suggest you import the whole logic from `libadalang_standalone.anod`: not only we’ll be sure that it will work on all supported platforms, but this will also enable us to remove that logic from there (and keep it in only one place).

make.py
Roldak: Well, the whole idea of this section is to put `$ORIGIN` as rpath, so first prevent gprbuild from passing its own rpaths using `-R` and then set the `LD_RUN_PATH` environment variable to `$ORIGIN` to make sure the linker uses it.

make.py
Roldak: No this won't work if `$ORIGIN` is not in the rpath as the dynamic loader won't look for shared libs in the current dir.

make.py
Roldak: Ah this is nice!

make.py
Roldak: Actually `*str` is the correct way from what I can see from Python editing tools (e.g. PyCharm complains with `list[str]` but typechecks correctly with `*str`)

langkit/dsl_unparse.py
pmderodat: There is no docstring, so I assume this was marked as resolved by mistake. ;-)

pmderodat:
To be honest I’m not sure https://github.com/AdaCore/langkit/pull/275/commits/26e2d2e453d5a1d188f227398469d64f5a6fe251 is a good idea: if we can’t import libpythonlang, we can’t unparse comments, and thus tests will likely fail in the future, as dsl_unparse won’t be able to unparse the comments required in the baseline. Can’t we instead consider that libpythonlang is now a real Langkit dependency?

langkit/dsl_unparse.py
Roldak: Mh no, there is a docstring! (https://github.com/AdaCore/langkit/pull/275/files#diff-b1a19ccbccaa0cda7fafe777e45fc165R83 -> ctrl+f `class_from_location`)

Roldak:
Okay, I can remove this commit.

langkit/dsl_unparse.py
pmderodat: Ah, I see: the comments were added in a later commit: I had the commit view only when I noticed this. Sorry for the noise!

PR #277, opened 2020-01-13
Work on lkt parser & unparser

raph-amiard:


langkit/compiled_types.py
```diff
@@ -1594,21 +1594,34 @@ def __init__(self, name, type, is_artificial=False, default_value=None,
...
                     or AbstractVariable(name, type, source_name=name))
         self.is_artificial = is_artificial
-        self.default_value = default_value
+
+        # Make sure that, if present, the default value is a compile-time known
+        # constant.
+
+        self.abstract_default_value = default_value
+        self.default_value = None
+        if default_value:
```
pmderodat: Given that `default_value` could be `False`, I think this should be `if default_value is not None`, what do you think?

raph-amiard: Gah, indeed, good catch, and really a great explanation of why I can't wait to get rid of AbstractExpressions

pmderodat: Sure! Needless to say: you’re not the only one!

PR #279, opened 2020-01-15
Add DSL features for struct types

pmderodat:


testsuite/tests/properties/invalid_struct_update/expected_concrete_syntax.lkt
```diff
@@ -0,0 +1,18 @@
+grammar None_grammar {
+    main_rule <- list*(Example(@Example))
+
+}
+
+class FooNode {
+
+    @export fun increment (kv : KV): KV = <StructUpdate>
```
raph-amiard: Would be cool if you added the syntax for it at the same time (in the unparser at least) if you feel up to it :)

pmderodat: Good idea. I’ll try!

langkit/compiled_types.py
```diff
@@ -1619,25 +1619,22 @@ def __init__(self, name, type, is_artificial=False, default_value=None,
...
-        self.default_value = None
-        if default_value is not None:
-            default_value = unsugar(default_value)
-            default_value.prepare()
-            self.default_value = construct_compile_time_known(
-                default_value, type
-            )
+        self.default_value = (
+            None
+            if default_value is None else
```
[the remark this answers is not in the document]
pmderodat: I think I remember we discussed style for this construct (leading to this formatting), years ago, but of course I can’t find the conversation, and it’s not tracked in our coding style document of course. :-/ Anyway, let’s discuss this live when we have a chance!

testsuite/tests/properties/invalid_struct_update/expected_concrete_syntax.lkt
raph-amiard: Thanks!

langkit/compiled_types.py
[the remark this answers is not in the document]
raph-amiard: Yes, sounds good. TBH what I would like to discuss live is use black and end once and for all python style discussions :)

PR #283, opened 2020-01-17
Expand concrete syntax

raph-amiard:


testsuite/tests/stylechecks/test.out
```diff
@@ -0,0 +1 @@
+testsuite/tests/contrib/lkt_parse_1/test.py:35:36: W291 trailing whitespace
```
pmderodat: Please remove the trailing whitespace

PR #285, opened 2020-01-21
Make a standalone test for Python grammar concrete syntax testing 

pmderodat:


testsuite/tests/properties/full_sloc_image/test.py
```diff
@@ -1,28 +1,23 @@
 """
-This tests that the Langkit based python parser can be compiled and can parse a
-simple python program without errors.
-
-TODO: This is hackish and minimalistic. Since Langkit's python parser is a
-fully fledged Langkit generated lib, it would make sense to reuse Libadalang's
-parser test driver.
+Test that the "full_sloc_image" worrks as expected.
```
Roldak: works*

pmderodat: Thanks! Now fixed.

PR #316, opened 2020-03-09
Minor: clean build space for lib*lang before building

raph-amiard:
This ensures that if the version of GNAT has changed, we won't get an
uncaught exception from out of date libraries.

### DIFF
--- a/scripts/manage.sh
+++ b/scripts/manage.sh
@@ -5,12 +5,26 @@ LANGKIT_ROOT=`dirname $0`"/.."
 cd "$LANGKIT_ROOT"
 
 make () {
+
+    # We need to clean the build space of the langkit libraries we depend
+    # upon before we run langkit again. Else, if we installed a newer version
+    # of GNAT since those libraries were built, what will happen is that:
+    #
+    # 1. Langkit will try loading them.
+    # 2. This will cause an uncaught exception trying to load some so from the
+    #    compiler, preventing langkit to run.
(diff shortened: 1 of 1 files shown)


scripts/manage.sh
```diff
@@ -7,10 +7,12 @@ cd "$LANGKIT_ROOT"
 make () {
     (
         cd "contrib/python"
+        rm build -rf
```
pmderodat: Can you please add comments to explain why we need this? In particular, I’m curious why the mechanism that makes GPRbuild rebuild files when the runtime (and thus the compiler) has changed is not enough. Do you think this is a GPRbuild bug? Also, what is this “uncaught exception” you are referring to?

raph-amiard: Well personally I think the commit message is enough, but if it can put your mind at ease, I did say what I said in the commit message here again in a comment.

I'm not sure how much clearer I can be, so tell me if you still don't understand/need more precisions.

pmderodat: It’s more about saving the trouble for future committers of looking in the commit history to understand where these comes from than putting my mind at ease. :-) The comments made crystal clear why removing the build directories is required, thank you very much.

raph-amiard: OK cool!

raph-amiard: Can you please re-approve then ? ;)

pmderodat: Sure! Sorry, I forgot that you pushed new changes and thus needed that…

PR #326, opened 2020-03-20
Further work on lkt semantics

raph-amiard:


contrib/lkt/language/parser.py
```diff
@@ -1923,28 +1949,37 @@ class GenericInstantiation(Expr):
...
+        )
+    )
 
     @langkit_property(public=True)
-    def designated_type():
+    def instantiated_decl():
         """
         Get the type designated by this instantiation.
 
         NOTE: for the moment we only have generic types, so that's enough. If
```
Roldak: Remove the note?

contrib/lkt/language/parser.py
```diff
@@ -1862,16 +1901,187 @@ def check_correctness():
...
+                .concat(S("` with `")).concat(expected_type.full_name)
+                .concat(S("`"))
+            )),
+
+            # Inference returned an error with message -> propagate it
+            Not(unsorted_actuals.error.is_null),
+            unsorted_actuals,
+
+            # We don't have all the actuals needed to infer -> return an error
+            unsorted_actuals.result.length
```
Roldak: It seems to me like this check is too soft. Consider for example:
```
def mkPair[T](x: T) -> Pair[T, T]

def foo(): Pair[Int, Bool] = mkPair(1)
```
Am I right that `infer_actuals_impl` would return `[T -> Int, T -> Bool]` here? So, this check would pass and further down in `check_infer_actuals` we will take the first actual for `T`, so `Int`, meaning the whole thing will typecheck "successfully".

So shouldn't we also check that the mappings returned by `infer_actuals_impl` contain no incoherent binding?

contrib/lkt/language/parser.py
raph-amiard: Good catch

contrib/lkt/language/parser.py
raph-amiard: That's correct, there is a missing check, good catch

I'll implement it in a further PR, since globally WIP anyway

PR #327, opened 2020-03-20
T218-020: Introduce Iterator Type

Roldak:


langkit/compile_context.py
```diff
@@ -2116,13 +2128,16 @@ def dependencies(typ):
         # by accident.
         struct_types = CompiledTypeRepo.struct_types
         array_types = CompiledTypeRepo.array_types
+        iterator_types = CompiledTypeRepo.iterator_types
         CompiledTypeRepo.struct_types = None
         CompiledTypeRepo.array_types = None
+        CompiledTypeRepo.iterator_types = None
 
         # Sort the struct and array types by dependency order
```
pmderodat: Please update the comment. No need to extend the enumeration: just say for instance: “Sort all composite types by dependency order”.

langkit/compiled_types.py
```diff
@@ -3626,6 +3653,113 @@ def require_hash_function(self):
...
 
 
+class IteratorType(CompiledType):
+    """
+    Base class for iterator types.
+    """
+    def __init__(self, name, element_type):
+        # For now, we expose all iterator types since their only purpose is
+        # to be used in the different public APIs and they don't even have a
+        # DSL API.
```
pmderodat: This is valid only if the `element_type` can be exposed itself, and we don’t know that at this point. I suggest to do as for arrays: consider it’s not exposed at construction, and enhance the `expose_public_api_types` pass to enable public API exposition when needed.

Besides, our ultimate goal is to use iterators internally, so I’m not sure we want to set in store that iterators are only for exposed types. If we do this nonetheless, please add a testcase that shows that we reject iterators on, say, lexical envs (which can’t be exposed).

langkit/compiled_types.py
```diff
@@ -3626,6 +3653,113 @@ def require_hash_function(self):
...
+
+        See StructType.emit_c_type.
+
+        :rtype: bool
+        """
+        return (not self.element_type.is_struct_type or
+                self.element_type.emit_c_type)
+
+    @property
+    def element_array_type_name(self):
```
pmderodat: IIUC, this is not an overriding method: please document it

langkit/compiled_types.py
```diff
@@ -3626,6 +3653,113 @@ def require_hash_function(self):
...
+        """
+        Name of the Ada iterator type.
+
+        :rtype: names.Name
+        """
+        return (self.api_name
+                if self.is_string_type else
+                (names.Name('Internal') +
+                    self.element_type.name +
+                    names.Name('Iterator')))
```
pmderodat: While this special treatment makes sense in `ArrayType.array_type_name` (we want to use `Wide_Wide_String` instead of declaring our own `Wide_Wide_Character` array), I see no reason for iterator not to behave similarly. Can you add a comment to explain this if it’s indeed required, or always return the `Internal_*_Iterator` type name?

langkit/expressions/collections.py
```diff
@@ -747,6 +747,25 @@ def collection_get(self, collection, index, or_null):
     return result
 
 
+@auto_attr
+def to_iterator(self, array):
```
pmderodat: Please document this

langkit/expressions/collections.py
```diff
@@ -747,6 +747,25 @@ def collection_get(self, collection, index, or_null):
...
+    array_expr = construct(array)
+    check_source_language(
+        array_expr.type.is_array_type,
+        '.at prefix must be an array: got {} instead'.format(
```
pmderodat: Please update the error message, and add a testcase for it (`properties/invalid_iterator`, to be similar to what’s done for other expressions)

langkit/expressions/collections.py
```diff
@@ -747,6 +747,25 @@ def collection_get(self, collection, index, or_null):
...
+            array_expr.type.dsl_name
+        )
+    )
+    elem_type = array_expr.type.element_type
+    result_type = elem_type.iterator
+
+    result = CallExpr('To_Iterator_Result', 'Iterator_From_Array', result_type,
+                      [array_expr])
+
+    result.abstract_expr = self
```
pmderodat: Can you move this to a CallExpr constructor argument?

langkit/ocaml_api.py
```diff
@@ -150,6 +151,11 @@ def array_wrapper(self, array_type):
                 if array_type.element_type.is_entity_type
                 else array_type).api_name.camel
 
+    def iterator_wrapper(self, iterator_type):
+        return (ct.T.entity.array
```
pmderodat: I guess this is a typo

langkit/templates/c_api/iterator_types_c.mako
```diff
@@ -0,0 +1,27 @@
...
+${c_doc(cls)}
+typedef void* ${type_name};
+</%def>
+
+<%def name="decl(cls)">
+
+<% type_name = cls.c_type(capi).name %>
+
+/* Set the next value of the iterator in the given element pointer.
+   Return 1 if successful, otherwise 0.  */
```
pmderodat: These docs are supposed to be shared in `langkit.documentation`: could you move them and use `${c_doc()}` to replace them where needed?

langkit/templates/c_api/pkg_main_spec_ada.mako
```diff
@@ -520,6 +521,16 @@ private package ${ada_lib_name}.Implementation.C is
       % endif
    % endfor
 
+   --------------------
+   -- Iterator types --
+   --------------------
+
```
pmderodat: I suggest to insert here documentation that will be common to all iterator types in this section: describe what iterators are, that these types have ref-counting reference semantics, and of course describe safety aspects once they are implemented.

pmderodat: … also document the same way the Next functions

langkit/templates/iterator_types_ada.mako
```diff
@@ -0,0 +1,187 @@
...
+  overriding procedure Adjust (T : in out ${cls.api_name}) is
+  begin
+     Inc_Ref (T.Internal_Iterator);
+  end Adjust;
+
+  --------------
+  -- Finalize --
+  --------------
+
+  procedure Finalize   (T : in out ${cls.api_name}) is
```
pmderodat: Extra spaces

langkit/templates/iterator_types_ada.mako
```diff
@@ -0,0 +1,187 @@
...
+</%def>
+
+<%def name="decl(cls)">
+
+   <% elt_type = cls.element_type.name %>
+
+   type ${cls.iterator_type_name} is record
+      Ref_Count : Integer;
+      --  Negative values are interpreted as "always living singleton".
+      --  Non-negative values have the usual ref-counting semantics.
```
pmderodat: IIUC we never need always living singletons for iterators. If so, can you update the comment accordingly?

langkit/templates/iterator_types_ada.mako
```diff
@@ -0,0 +1,187 @@
...
+
+  procedure Free is new Ada.Unchecked_Deallocation
+    (${cls.iterator_type_name}, ${cls.name});
+</%def>
+
+<%def name="body(cls)">
+
+   <% elt_type = cls.element_type.name %>
+
+   function Iterator_From_Array
```
pmderodat: Please add comment boxes

langkit/templates/python_api/iterator_types_py.mako
```diff
@@ -0,0 +1,87 @@
...
+    __slots__ = ('c_value',)
+
+    def __init__(self, c_value):
+        self.c_value = c_value
+
+    def __repr__(self):
+        return '<{}>'.format(type(self).__name__)
+
+    def clear(self):
+        self.c_value = None
```
pmderodat: This method is for internal use: please add a `_` prefix (same for wrap and wrap_iter IIUC)

langkit/templates/python_api/iterator_types_py.mako
```diff
@@ -0,0 +1,87 @@
...
+
+    def __del__(self):
+        self.dec_ref(self.c_value)
+        self.clear()
+
+    @classmethod
+    def wrap(cls, c_value):
+        res = cls(c_value)
+        # inc_ref?
```
pmderodat: I don’t think you need an `inc_ref` here: unless I’m missing something, the only use case for `wrap` when handling a value returned from a C function: in these cases, the function returns an ownership share, and the wrap method is here only to convert this value (and there will be no direct user for this value afterwards). It’s worth adding as a comment IMO.

langkit/templates/python_api/iterator_types_py.mako
```diff
@@ -0,0 +1,87 @@
...
+<%def name="decl(cls)">
+
+<%
+    element_type = cls.element_type
+    c_element_type = pyapi.c_type(element_type)
+%>
+
+class ${cls.py_converter}(_BaseIterator):
+    """
+    Wrapper class for iterators of ${cls.element_type.name}.
```
pmderodat: This documentation is for users, so I suggest just writing: `Iterator over ${cls.…}`

langkit/templates/python_api/iterator_types_py.mako
```diff
@@ -0,0 +1,87 @@
...
+
+    This class is not meant to be directly instantiated: it is only used to
+    convert values that various methods take/return.
```
pmderodat: Given that iterators are only returned by properties, I suggest rewording:

> This class is not meant to be directly instantiated: it is only used for the return values of properties returning iterators.

langkit/templates/python_api/iterator_types_py.mako
pmderodat: Likewise for `c_value`

langkit/templates/python_api/iterator_types_py.mako
```diff
@@ -0,0 +1,87 @@
...
+    def wrap_item(item):
+        return ${pyapi.wrap_value('item', element_type,
+                                  from_field_access=False)}
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        x = self.c_element_type()
+        if self.get_next(self.c_value, ctypes.pointer(x)):
```
pmderodat: The doc for ctypes recommends to use `ctypes.byref` instead of `.pointer` when just passing a temporary reference to a function call. (https://docs.python.org/2/library/ctypes.html#passing-pointers-or-passing-parameters-by-reference)

langkit/templates/python_api/iterator_types_py.mako
```diff
@@ -0,0 +1,87 @@
...
+            return self.wrap_item(x)
+        raise StopIteration
+
+    # For Python2 compatibility
+    next = __next__
```
pmderodat: Could you move this to the base class?

langkit/templates/python_api/iterator_types_py.mako
pmderodat: and all the C types and function wrappers in subclasses :-)

testsuite/tests/iterators/int_iterator/main.adb
```diff
@@ -0,0 +1,33 @@
...
+   U   : constant Analysis_Unit := Ctx.Get_From_Buffer
+     (Filename => "main.txt",
+      Buffer   => "example");
+begin
+   Put_Line ("main.adb: Running...");
+   if U.Has_Diagnostics then
+      raise Program_Error;
+   end if;
+
+   --  Through array
```
pmderodat: I suggest to turn these comments into calls to `Put_Line` for `test.out` readability

testsuite/tests/iterators/int_iterator/main.py
```diff
@@ -0,0 +1,25 @@
...
+        print(d)
+    sys.exit(1)
+
+print(u.root.p_values_array)
+it = u.root.p_values_iterator
+try:
+    while True:
+        print(next(it))
+except StopIteration:
+    pass
```
pmderodat: Why not using a `for` loop instead of emulating the iterator protocol?

langkit/expressions/base.py
```diff
@@ -2905,6 +2905,8 @@ def construct_static(elements, array_type, abstract_expr=None):
             )
```
pmderodat: Can you add to the commit message the reason why you need this? I guess something else breaks without this patch, but it would be useful to know what

langkit/compile_context.py
```diff
@@ -2124,6 +2124,10 @@ def dependencies(typ):
             # Filter types that are relevant for dependency analysis
             return [t for t in result if t.is_struct_type or t.is_array_type]
 
+        # Force creation of the entity iterator type, so that it is available
+        # later during public APIs generation.
+        CompiledTypeRepo.root_grammar_class.entity.iterator
+
```
pmderodat: IIUC, this is needed if and only if there is at least one iterator type for any entity type. So what about forcing this creation in the `IteratorType` constructor when called on a non-root entity type?

langkit/templates/iterator_types_ada.mako
```diff
@@ -5,7 +5,10 @@
 
   function Next
     (I : ${cls.api_name};
-     E : out ${cls.element_type.api_name}) return Boolean;
+     E : out ${"{}{}".format(
+         cls.element_type.api_name,
+         "'Class" if cls.element_type.is_entity_type else ''
+     )}) return Boolean;
```
pmderodat: This looks wrong to me: assuming that B derives from A, if we have an iterator on A, we don’t want to allow users to pass a B actual (which matches A'Class).

langkit/compile_context.py
```diff
@@ -2124,6 +2124,10 @@ def dependencies(typ):
             # Filter types that are relevant for dependency analysis
```
pmderodat: After I’ve read this commit, I failed to understand what it really does. What is shared exactly?

langkit/templates/pkg_implementation_body_ada.mako
```diff
@@ -1895,8 +1895,9 @@ package body ${ada_lib_name}.Implementation is
             end if;
 
             --  Call recursively on children
-            for C of ${root_node_array.array_type_name}'(Children (Node)) loop
-               Result := Populate_Internal (C, Node.Self_Env) or else Result;
+            for I in First_Child_Index (Node) .. Last_Child_Index (Node) loop
+               Result := Populate_Internal
+                 (Child (Node, I), Node.Self_Env) or else Result;
```
pmderodat: Interesting. Could you create a dedicated testcase in `lexical_envs/` for this?

testsuite/tests/iterators/big_iterator/test.py
```diff
@@ -0,0 +1,38 @@
+"""
+Test iterating over a huge number of elements.
+
+This shows that an iterator should be used in place of an array when one wants
+to return a huge number of elements, as calling ``p_get_all_examples`` from
+the public APIs crashes in this instance.
```
pmderodat: When possible, I think we should avoid committing huge files (this is for practical reasons: keep the git repo small to keep it convenient to handle).

Here, we just need a big array in the DSL. A huge list of parsing nodes is one way to do it (with this huge file con), but another way would be to have a property that takes an array and that returns an iterator over that array, and do from the Python script: `u.root.p_iter_array(range(10**9))` (likewise in Ada)

langkit/dsl_unparse.py
```diff
@@ -973,6 +973,8 @@ def bases(typ):
             return type.raw_name.camel
```
pmderodat: Given that the previous implementation, present in this PR, was wrong, what about squashing this commit?

PR #333, opened 2020-04-02
 T103-024: Fix bug in precise element types computation.

Roldak:


testsuite/tests/properties/precise_synth_fields/test.py
```diff
@@ -24,6 +27,21 @@ class ParsedHolder(AbstractHolder):
     pass
 
 
+class HolderList(AbstractHolder.list):
+    pass
+
+
+@abstract
+class AbstractManyHolder(FooNode):
```
pmderodat, line 35: ManyHolderS ;-)

pmderodat: As disucssed live, debatable as it’s also this class that is a holder. Anyway.

PR #334, opened 2020-04-02
Work on enum members in structs

raph-amiard:


langkit/compiled_types.py
```diff
@@ -3649,8 +3647,22 @@ def __init__(self, name, location, doc, value_names):
         :type: dict[names.Name: EnumValue]
         """
 
+        self.default_val_name = default_val_name
+        """
+        Name of the default value for this enum, if any
```
pmderodat, line 3652: Missing trailing dot

langkit/compiled_types.py
```diff
@@ -3649,8 +3647,22 @@ def __init__(self, name, location, doc, value_names):
...
+        :type: names.Name|None
+        """
+
         CompiledTypeRepo.enum_types.append(self)
 
+        super(EnumType, self).__init__(
+            name, location, doc, is_ptr=False, exposed=True,
+            null_allowed=default_val_name is not None,
+            nullexpr=self.values_dict[default_val_name].ada_name
+            if default_val_name is not None else None,
```
pmderodat, line 3663: Can you reindent the `if` so it’s clear that it applies to `self.values_dict[…`?

langkit/dsl.py
```diff
@@ -958,6 +959,15 @@ def __new__(mcs, name, bases, dct):
                 )
                 value._name = names.Name.from_lower(key)
                 values.append(value)
+
+                # If this is the default value for this enum type, store it
+                if value.is_default_val:
+                    check_source_language(
+                        default_val_name is None,
+                        'Only one default value is allowed'
+                    )
```
pmderodat, line 968: Can you add a testcase that show that this check works when there are more than one default value? `properties/invalid_enum_dflt` for instance, just calling emit_and_print_errors, like other such validation tests.

testsuite/tests/properties/struct_enum_member/test.py
```diff
@@ -0,0 +1,73 @@
+"""
+Test basic features for enumeration types.
```
pmderodat, line 2: Please update the docstring

testsuite/tests/properties/struct_enum_member/main.py
```diff
@@ -0,0 +1,29 @@
...
+for d in r:
+    print('{}.with_kind = {}'.format(d, d.p_with_kind))
+
+for value in (None, 1, 'blah', 'func'):
+    try:
+        result = r.p_identity(value)
+    except Exception as exc:
+        result = '<{}: {}>'.format(type(exc).__name__, exc)
+    else:
+        result = repr(result)
```
pmderodat, line 27: This loop looks unused: can you remove it?

langkit/dsl.py
raph-amiard: I'll add it if you want, however I'm not sure of the value at this stage:

1. This check is going to bounce when we transition to the concrete syntax, probably replaced by a check in lkt.
2. The test I add will need to be rewritten manually, because we cannot unparse those, because those checks abort before we can even emit the .lkt file

pmderodat: I think the value of this one is the same as for the other validation checks: make sure that the feature works as expected now, and help to cover this case in the future (especially when migrating to Lkt). This test is supposed to be very easy to write, and so porting this one (and all other such tests) manually should be easy enough.

This is similar to the invalid grammar tests we had to transition to Lkt: it took 1.5 days to port them all, but having them has value to validate the Lkt implementation (and did help us find bugs and missing features!).

raph-amiard: Ok, will go ahead then

pmderodat: Thanks!

PR #336, opened 2020-04-22
T421-021: Return foreign results in Env.get(*)

raph-amiard:


langkit/support/langkit_support-lexical_env.adb
```diff
@@ -659,12 +659,24 @@ package body Langkit_Support.Lexical_Env is
...
+
+               --  Append internal nodes
+               for I in reverse I_Nodes.First_Index .. I_Nodes.Last_Index loop
+                  Append_Result
+                    (I_Nodes.Get (I), Metadata, Current_Rebindings,
+                     From_Rebound);
+               end loop;
+
+               --  Append foreign nodes
+               for C in reverse F_Nodes.Iterate loop
```
pmderodat, line 676: I suggest to remove the `reverse` here, to be consistent with the handling of foreign nodes in other lookups.

raph-amiard: Done! and added a testcase

PR #342, opened 2020-05-05
Some lkt work

raph-amiard:
* Split lkt tests
* Some nameres
* Don't emit `EnvAssoc`


contrib/lkt/language/parser.py
```diff
@@ -2078,9 +2082,14 @@ def called_decl():
...
-            ._.get_type._.get_fun('__call__')
-            .then(lambda a: a, default_val=refd_decl)
+            refd_decl.match(
+                # Don't try to look for __call__ on function decls
+                lambda _=FunDecl: refd_decl,
+                lambda v=BaseValDecl:
+                v._.get_type._.get_fun('__call__')
+                .then(lambda a: a, default_val=refd_decl),
+                lambda _: refd_decl,
+            )
```
pmderodat, line 2127: Can you add a testcase for the cases where we had crashes and don’t anymore?

raph-amiard: Erf I had one when I did this... Don't know where I put it... I'll manage!

PR #343, opened 2020-05-14
OCaml bindings: fix a bug in types dependency tracking

pmderodat:


langkit/ocaml_api.py
```diff
@@ -16,19 +16,46 @@ class AnalysisContextType(object):
...
+        if typ.is_entity_type:
+            # Expose all entity types as the root entity type
+            result = T.entity
+        elif typ.is_array_type:
+            result = self.actual_c_type(typ.element_type).array
+        else:
+            result = typ
+
+        if not isinstance(result, self.AnalysisContextType):
+            result = resolve_type(result)
```
raph-amiard, line 42: Why do you have a special case on `self.AnalysisContextType` ? This code is not very clear, and anyway `resolve_type` is made to sort out that kind of stuff by itself, why have a condition ?

pmderodat: AnalysisContextType is not a CompiledType (see above in this module): it’s a hack in the ocaml bindings generator to properly order type declarations, including analysis contexts. So resolve_type does not work for it.

raph-amiard: Ok it probably deserves a comment then!

pmderodat: Fair enough. Comment added, thanks.

langkit/ocaml_api.py
```diff
@@ -16,19 +16,49 @@ class AnalysisContextType(object):
...
+            result = T.entity
+        elif typ.is_array_type:
+            result = self.actual_c_type(typ.element_type).array
+        else:
+            result = typ
+
+        # Make sure we get a CompiledType instance, except for
+        # AnalysisContextType, which is a placeholder (see its declaration
+        # above).
+        if not isinstance(result, self.AnalysisContextType):
```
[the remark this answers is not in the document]
pmderodat: Good point, I’ll adjust.

langkit/ocaml_api.py
```diff
@@ -16,19 +16,49 @@ class AnalysisContextType(object):
...
         Adds the dependency dep to the type typ in the type_graph.
         """
-        if typ not in self.type_graph:
-            self.type_graph[typ] = []
-
-        if dep not in self.type_graph:
-            self.type_graph[dep] = []
+        # Add a dependency both to the actual C type (for C structs order) and
+        # to the given type (for OCaml modules order).
+        for dep in [dep, self.actual_c_type(dep)]:
```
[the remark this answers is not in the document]
pmderodat: Yes, I was hoping the comment was clarifying this: while `self.actual_c_type(dep)` is necessary for the ordering of C structs (https://github.com/AdaCore/langkit/blob/627e4094d0b9211557117fdc668e0a396fc9b06d/langkit/templates/ocaml_api/module_ocaml.mako#L510), `dep` is still needed for the ordering of OCaml modules (https://github.com/AdaCore/langkit/blob/627e4094d0b9211557117fdc668e0a396fc9b06d/langkit/templates/ocaml_api/module_ocaml.mako#L835).

If you don’t think the comment is helpful, would you have a suggestion to enhance it?

PR #350, opened 2020-05-29
More unparsing & type resolution work on LKT

raph-amiard:


contrib/lkt/language/parser.py
```diff
@@ -1246,6 +1295,11 @@ def type_base_scope():
 
     call_scope = Property(Entity.type_scope)
 
+    is_generic = Property(
+        Entity.is_a(T.InstantiatedGenericType), public=True,
+        doc="Returns whether this type is an instantiated generic type"
```
pmderodat, line 1320: This should be “return”, and this lacks a trailing dot.

contrib/lkt/language/lexer.py
```diff
@@ -158,9 +160,10 @@ class Token(LexerToken):
     # Numbers
     (Pattern('[0-9]+'),    Token.Number),
 
-    # Strings
-    (Pattern('{STRING_SQ}|{STRING_DBQ}'), Token.String),
-    (Pattern('[a-zA-Z]{STRING_DBQ}|[a-zA-Z]{STRING_DBQ}'), Token.PString),
+    # Strings & chars
+    (Pattern('{STRING_DBQ}|{STRING_SQ}'),         Token.String),
+    (Pattern('[a-zA-Z]{STRING_DBQ}'), Token.PString),
+    (Pattern('{CHAR_LIT}'),           Token.Char),
```
pmderodat, line 166: Looks like `'a'` can now be lexed both as a single-quoted string literal, and as a charactel literal. What about we get rid of single-quoted string literals?

contrib/lkt/language/parser.py
raph-amiard: As discussed live, I have no intention of following PEP 257 here or elsewhere, and I'll add the trailing dot :slightly_smiling_face:

contrib/lkt/language/lexer.py
raph-amiard: Yes, thanks for catching that, I actually thought I did get rid of them!

PR #352, opened 2020-05-29
T529-011 Implement BigInteger using zarith in OCaml bindings

danielmercier:


testsuite/tests/ocaml_api/general/main.ml
```diff
@@ -406,7 +406,10 @@ let test_big_int () =
   let u = AnalysisContext.get_from_file ctx "foo.txt" in
   print_exit_if_diags u ;
   let root = root_exn u in
-  Format.printf "@[<v>int_double 42: %d@ @]" (FooNode.p_int_double root 42) ;
+  Format.printf "@[<v>int_double 42: %s@ @]"
+    (Z.to_string (FooNode.p_int_double root (Z.of_int 42))) ;
+  Format.printf "@[<v>int_double 2147483647: %s@ @]"
+    (Z.to_string (FooNode.p_int_double root (Z.of_int 2147483647))) ;
```
pmderodat, line 412: This number isn’t the same as in the baseline: what’s going on? ;-) In any case, could you try with a number that clearly exceeds 64 bits? Maybe 10**100?

danielmercier: This is to translate an integer to a big integer. The number here is an integer, but is then doubled by the property p_int_double, thus, it is not an integer afterwards.

In any case, it those not exceeds 64 bits, I can change if you really want.

pmderodat: Ah, I see, sorry! Yeah, a very big integer sounds nice: I assume it’s easy using zarith to build 10**100?

PR #355, opened 2020-06-01
T511-017: make stylechecks work with pycodestyle

raph-amiard:


langkit/lexer/__init__.py
```diff
@@ -879,7 +879,7 @@ def __init__(self, match_length, *alts):
             self.location = extract_library_location()
             self.match_length = match_length
 
-            for i, alt in enumerate(alts):
+            for _, alt in enumerate(alts):
```
pmderodat, line 882: Let’s remove the call to enumerate altogether

langkit/stylechecks/__init__.py
```diff
@@ -452,25 +452,36 @@ def check(self, report, filename, content, parse):
         self.custom_check(report, filename, content, parse)
         if os.path.exists(filename):
             self.pep8_check(report, filename)
-            self.pyflakes_check(report, filename, content)
+            # self.pyflakes_check(report, filename, content)
```
pmderodat, line 455: Is this permanent? Can you explain?

langkit/lexer/__init__.py
raph-amiard: Right ... :derp:

langkit/stylechecks/__init__.py
raph-amiard: It's a mistake ... Or the reason code reviews exists :) Thanks !

PR #372, opened 2020-06-17
More mypy checks & refactorings

pmderodat:


langkit/lkt_lowering.py
```diff
@@ -984,6 +1004,35 @@ def lower(rule: L.GrammarExpr) -> Optional[Parser]:
         grammar._add_rule(name, lower(rule))
 
 
+EnvType = Dict[L.BaseValDecl, AbstractVariable]
+
+
+class Environment:
+    """
+    Helper to track the association between Lkt definitions and our
```
raph-amiard, line 1012: This is not crystal clear :slightly_smiling_face: since you also have several dicts for types. I think you should say this is for local variables in properties.

Also I wonder if this mechanism couldn't be used for everything, with a root env for types ... But I do'nt know, it might not be practical.

pmderodat: As discussed live: I’ll refactor the dicts for type to have a simpler types lowering system, and I’ll replace this Environment object with a simple dict for local variables: we don’t need to handle scopes as Lkt is already processing them correctly.

PR #373, opened 2020-06-17
Work on langkit diagnostics

raph-amiard:


langkit/diagnostics.py
```diff
@@ -65,77 +68,46 @@ def set_style(cls, style):
...
-        loc = self
-        while loc is not None:
-            loc_str = '{}:{}'.format(P.basename(loc.file), loc.line)
-            if loc.column:
-                loc_str += ':{}'.format(loc.column)
-            stack.append(loc_str)
-            loc = loc.previous_in_callstack
-
-        return '[{}]'.format(', '.join(stack))
+        return "{}:{}:{}".format(self.file, self.line, self.column)
```
pmderodat: GNU tools don’t print the column number when it is unknown, and that’s what the previous implementation did. Was this change intentional?

langkit/diagnostics.py
```diff
@@ -6,7 +6,9 @@
 import os.path as P
 import sys
 import traceback
-from typing import NoReturn
+from typing import NoReturn, Optional
+
+import liblktlang as L
```
pmderodat: This breaks the bootstrap on this commit. Could you move the introduction of the `try/except ImportError` block in this commit?

langkit/diagnostics.py
```diff
@@ -100,28 +102,35 @@ class Location:
...
 
-    @property
-    def gnu_style_repr(self) -> str:
+    lkt_unit: Optional[L.AnalysisUnit] = field(default=None)
+
+    def gnu_style_repr(self, relative: bool = True) -> str:
         """
         Return a GNU style representation for this Location, in the form::
 
             file:line:column
```
pmderodat, line 118: Please document the `relative` argument

langkit/diagnostics.py
```diff
@@ -607,3 +609,106 @@ def errors_checkpoint():
...
+    # Print the source listing
+    if location.lkt_unit is not None:
+        print(source_listing(location))
+
+
+def print_error_from_sem_result(sem_result: L.SemanticResult):
+    """
+    Prints an error from an lkt semantic result.
+
+    TODO: This should be moved in langkit.diagnostics once we use LKT.
```
pmderodat, line 676: Obsolete comment ;-)

langkit/diagnostics.py
raph-amiard: No it's a mistake, good catch!

langkit/diagnostics.py
raph-amiard: Done

langkit/diagnostics.py
raph-amiard: Thanks :sweat_smile:

langkit/diagnostics.py
pmderodat: It’s not fixed, yet: we can see `:0: ` patterns in the rewritten baselines

PR #376, opened 2020-06-19
Lkt lowering: extend support for expressions

pmderodat:


langkit/lkt_lowering.py
```diff
@@ -248,9 +258,9 @@ class FlagAnnotationSpec(AnnotationSpec):
     """
     Convenience subclass for flags.
     """
-    def __init__(self, name: str):
+    def __init__(self, name: str, default_value: Any = False):
```
raph-amiard, line 261: I think this is hackish. A flag is either passed or not. Here you're trying to conflate a syntactic attribute (presence or not of the flag) into its underlying meaning (whether the property is public or not), except this info is already stored in its semantic state in `PropertyDef`, so no need to do it here. Also you're adding a param to a "flag" spec that doesn't make a lot of sense IMHO...

A simpler and cleaner design IMO is to just pass `annotations.export or None` at the single spot where you need to convert this info for PropertyDef.

langkit/lkt_lowering.py
```diff
@@ -1630,6 +1659,37 @@ def create_enum_node_alternatives(
...
+        value_names = []
+        for lit in decl.f_literals:
+            name = lit.f_syn_name.text
+            check_source_language(
+                name not in value_names,
+                'The "{}" literal is present twice'
+            )
+            value_names.append(name)
+
+        check_source_language(
```
raph-amiard, line 1724: Seems to me you'd go faster checking this as a whitelist: Only nodes can implement traits for the moment, and only built-in traits.

langkit/dsl_unparse.py
```diff
@@ -1068,11 +1068,24 @@ def emit_doc(doc):
...
+
+    if prop_for_walker.is_public:
         quals += "@export "
 
-    if prop.abstract:
-        quals += "abstract "
+    if not prop.constructed_expr:
+        quals += "@abstract "
+        if prop.abstract_runtime_check:
+            quals += "@abstract_stub "
```
raph-amiard, line 1088: I'm not fond of that name, even though I acknowledge the need to find something better than `abstract_runtime_check`. What about `runtime_error` or `not_implemented` ?

```ocaml
@not_implemented fun foo()
```

langkit/expressions/base.py
```diff
@@ -3212,6 +3212,21 @@ def __init__(self, expr, prefix, name=None, doc=None, public=None,
         :type: bool
         """
 
+        self.is_artificial_dispatcher = False
```
raph-amiard, line 3215: I remember feeling uneasy about this at some point when I worked on trying to unify the 17 fields we have for the names of properties, but anyway, for me this is really not so clear: 

All dispatchers are generated, insofar as the code is generated, so saying that some dispatchers "come from source" is very misleading IMO. What this means is that you didn't have to move the true body of the property to another property because there is no body to speak of.

Also it seems like the only use of this field is for your assertion in dsl_unparse ?

Anyway, I wonder if a simpler solution to that problem might just be running the unparser *before* those expansion passes ?

testsuite/tests/properties/entity_bind/expected_concrete_syntax.lkt
```diff
@@ -6,16 +6,12 @@ grammar foo_grammar {
 
 @abstract class FooNode implements Node {
 
-    @export fun prop (): Int
-
-    fun prop (): Int
+    @export @abstract @abstract_stub fun prop (): Int
```
raph-amiard, line 9: I think `@abstract_stub` or `@runtime_error` should replace `@abstracŧ` here ... Is that a mistake ?

langkit/lkt_lowering.py
```diff
@@ -1695,6 +1718,27 @@ def create_enum(self,
             value_names=[names.Name.from_lower(n) for n in value_names],
         )
 
+    def create_struct(self,
+                      decl: L.StructTypeDecl,
+                      annotations: StructAnnotations) -> StructType:
+        """
+        Create an StructType instance.
```
raph-amiard, line 1740: "Create *a"

langkit/lkt_lowering.py
pmderodat: I don’t understand your suggestion: how could we use a whitelist while no trait is allowed? The whitelist would be empty, right? In any case, I just realized that TokenNode should be legal for enum types, so I guess I should just remove this check.

pmderodat: Never mind, this is an enum type, not an enum node: no trait should be allowed.

testsuite/tests/properties/entity_bind/expected_concrete_syntax.lkt
pmderodat: No, that was intended, but I’m okay with using `@not_implemented` or `@runtime_error` to mean both, with a slight preference towards `@not_implemented`.

langkit/expressions/base.py
pmderodat: I’d prefer avoiding moving the dsl_unparse pass so that it’s still in the “emit” pass and thus so that we can rely on the fact that all compilation happened (i.e. all information is available).

As discussed live, the dispatch lowering pass is very tricky, and there is a mind-twisting reuse of PropertyDef objects, making it hard to think about object identity across the compilation pipeline. I’ll try to simplify this first and remove the need for this extra field.

raph-amiard: > As discussed live, the dispatch lowering pass is very tricky, and there is a mind-twisting reuse of PropertyDef objects, making it hard to think about object identity across the compilation pipeline. I’ll try to simplify this first and remove the need for this extra field.

Ok, that works for me, although it's definitely not urgent so keep that for later if you prefer!

testsuite/tests/properties/entity_bind/expected_concrete_syntax.lkt
raph-amiard: Ok, let's go with `@not_implemented` then!

langkit/lkt_lowering.py
pmderodat: As discussed live: let’s factorize the check for the presence/absence of traits to `resolve_type_decl` to avoid repeating it in all the `create_*` helpers.

testsuite/tests/properties/entity_bind/expected_concrete_syntax.lkt
pmderodat: Done thanks

PR #383, opened 2020-07-01
Rename Children_With_Trivia and make it return ghost nodes

pmderodat:


langkit/dsl_unparse.py
```diff
@@ -535,6 +535,7 @@ def var_name(var_expr, default="_"):
 
 def is_a(expr, *names):
     return any(expr.__class__.__name__ == n for n in names)
+expr_is_a = is_a
```
Roldak, line 538: why not rename `is_a`  altogether instead of introducing an alias? If we're keeping this anyway, maybe add a newline here?

pmderodat: Originally, I needed to call this one from `emit_expr`, which creates its own `is_a` function, so I just created this alias to workaround that. I thought that renaming it instead of creating an alias would force a huge refactoring, but actually there is only one call to this global is_a, so I’ve just done the renaming… Thanks!

Roldak: ok cool!

PR #387, opened 2020-07-16
Expose filename member for analysis unit in Ocaml bindings

leocreuse:
This PR exposes the filename attribute of the analysis units in the Ocaml bindings.
This also fixes memory leaks on the function token_kind_name in module Token.

### DIFF
--- a/langkit/templates/ocaml_api/module_ocaml.mako
+++ b/langkit/templates/ocaml_api/module_ocaml.mako
@@ -254,6 +254,19 @@ end
 
 % endfor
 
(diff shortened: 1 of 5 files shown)


pmderodat:
Hello Léo,

Thanks for your contribution, and in particular for taking care of testing your changes. Could you open an AdaCore ticket and add the customary `TN: XXXX-XXX` line to the commit message? With this, everything looks good to me.

leocreuse:
Hello PM, 
Done, thanks for the review

PR #388, opened 2020-07-16
Add typed lookup utility for Ocaml bindings

leocreuse:
This new utility lets a user lookup the deepest node of a given type
containing the specified source location. The returned node is of
the specified type if found, None otherwise.

### DIFF
--- a/langkit/templates/ocaml_api/module_ocaml.mako
+++ b/langkit/templates/ocaml_api/module_ocaml.mako
@@ -1134,6 +1134,24 @@ let ${ocaml_api.field_name(field)}
     let rec aux node = p node && for_all_fields aux node in
     aux (node :> ${root_entity_type})
 
+  let lookup_with_kind : type a. a node -> [< ${root_entity_type}] -> Sloc.t -> a option =
+    fun node_type node sloc ->
+      let lookup_res = lookup node sloc in
+      let rec aux : a node -> [< ${root_entity_type}] -> a option =
+        fun node_type node -> 
+        match node_type, node with
+      %for astnode in ctx.astnode_types:
+        | ${ocaml_api.node_name(astnode)}
+          , (#${ocaml_api.type_public_name(astnode)} as node) ->
+          Some node
+      %endfor
+        | _ -> (match parent node with
+                | Some parent_node -> aux node_type parent_node
+                | _ -> None) in
+    match lookup_res with
+      | Some node -> aux node_type node
+      | _ -> None
+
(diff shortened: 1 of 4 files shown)


testsuite/tests/ocaml_api/general/main.ml
```diff
@@ -217,6 +217,22 @@ let test_node () =
     (FooNode.for_all aux root) ;
   Format.printf "@[<v>============================@ @ @]"
 
+let test_lookup_with_kind () = 
+  Format.printf "@[<v>========== LOOKUP ==========@ @]" ;
+  let ctx = AnalysisContext.create () in
+  let u = AnalysisContext.get_from_file ~reparse:true ctx "foo.txt" in
+  print_exit_if_diags u ;
+  let root = root_exn u in
+  let lookup_1 = FooNode.lookup_with_kind Sequence root {Sloc.line=1; Sloc.column=25} in
```
danielmercier, line 226: SInce the goal of lookup_with_kind is to have a correct type for the returned node, I suggest to add a typing information for lookup_1. Something like:
let lookup_1 : Sequence.t option.

testsuite/tests/ocaml_api/general/main.ml
```diff
@@ -217,6 +217,22 @@ let test_node () =
...
+  (match lookup_1 with
```
danielmercier, line 227: It seems that you did not run ocamlformat on the code. I would prefer you do so.
We changed ocamlformat version recently so other parts of the code will change with a call to ocamlformat unfortunately, but this is fine

langkit/templates/ocaml_api/module_ocaml.mako
```diff
@@ -1145,6 +1163,7 @@ let ${ocaml_api.field_name(field)}
     | _ ->
         None
 
+
```
danielmercier, line 1166: You can remove this space

testsuite/tests/ocaml_api/general/main.ml
leocreuse: Done in  c076216

testsuite/tests/ocaml_api/general/main.ml
leocreuse: Done in b0daa20

langkit/templates/ocaml_api/module_ocaml.mako
leocreuse: Done in b02aa20

PR #390, opened 2020-07-28
Use Append_One instead of Append for containers

yakobowski:
Work for AI12-0339 will make this use of Append ambiguous, so use
Append_One to get forward compatible code.

TN: T722-036


pmderodat:
Append_One is an Ada 2020 feature, right? If so, we need to find a workaround, as LAL must build with the latest Community & FSF releases.

yakobowski:
Righ, I forgot about this... In that case I will use a qualified expression.

PR #393, opened 2020-08-10
Fix Windows build by using an executable with subprocess calls

LordAro:
Got the following error when trying to build on Windows

```
Traceback (most recent call last):
  File "manage.py", line 103, in <module>
    args.func(args, unknown_args)
  File "manage.py", line 35, in wrapper
    fn(args)
  File "manage.py", line 67, in make
    m1 = subprocess.Popen(
  File "C:\Python38\lib\subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\Python38\lib\subprocess.py", line 1308, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
OSError: [WinError 193] %1 is not a valid Win32 application
```

Windows requires an executable as the first argument, can't get away with relying on a shebang here!

Added the missing executables to the calls, and made one existing one more consistent.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -44,7 +44,10 @@ def setenv(args: Namespace) -> None:
     Setenv for langkit.
     """
     for cwd in (LKT_LIB_ROOT, PYTHON_LIB_ROOT):
-        subprocess.check_call(["./manage.py", "setenv"], cwd=cwd)
+        subprocess.check_call(
+            [sys.executable, "./manage.py", "setenv"],
+            cwd=cwd
+        )
 
 
 def make(args: Namespace) -> None:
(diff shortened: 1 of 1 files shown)


LordAro:
Curious - I found it in an MSYS2 environment as well. Though I was using "genuine" Windows Python, perhaps that's the difference?

pmderodat:
Ah, never mind: actually our production scripts directly call the `contrib/{python,lkt}/manage.py` scripts: Langkit’s toplevel `manage.py` script is really here for developpers’ convenience… and all our developpers happen to work on GNU/Linux boxes, hence the blind spot. ;-)

PR #394, opened 2020-08-12
T811-021 Add support for using GNATCOLL.Memory in main programs

setton:
Prep work for adding support for memory reporting in nameres.

### DIFF
--- a/langkit/templates/mains_project_file.mako
+++ b/langkit/templates/mains_project_file.mako
@@ -33,6 +33,10 @@ project Mains is
             --  profiling, for instance.
             for Default_Switches ("Ada") use ("-g", "-Ofast");
       end case;
+
+      --  A version of s-memory may be included for memory monitoring
+      --  with GNATCOLL.Memory: if so, it requires -gnatg.
+      for Switches ("s-memory.adb") use ("-g", "-O2", "-gnatpg");
    end Compiler;
 
 


langkit/templates/mains_project_file.mako
```diff
@@ -33,6 +33,10 @@ project Mains is
             --  profiling, for instance.
             for Default_Switches ("Ada") use ("-g", "-Ofast");
       end case;
+
+      --  A version of s-memory may be included for memory monitoring
+      --  with GNATCOLL.Memory: if so, it requires -gnatg.
```
pmderodat, line 38: The option is `-gnatpg`, no? (typo in the comment?)

Could you help me understanding how this should work, for future maintainance work? In order  for GPRbuild to rebuild `s-memory.adb`, users should copy its source in the generated project source directory?

setton: The way this works is that `GNATCOLL.Memory` requires you to provide your own `s-memory.adb`. See [gnatcoll-memory.ads](https://github.com/AdaCore/gnatcoll-core/blob/master/src/gnatcoll-memory.ads). 

`-gnatg` is what's strictly required so that the compiler will agree to compile GNAT units. The other options are here in the hope they help performance.

pmderodat: Thanks. Hm… the doc in gnatcoll-memory mentions gnatmake, so I’m not sure what’s the procedure you followed exactly to make it work.

setton: The addition of this line in the .gpr is what's needed to make it work in the gprbuild way.

pmderodat: Oh that’s convenient. Thanks!

PR #395, opened 2020-08-12
Add lkt_toolbox cmd line tool + some lkt type res

raph-amiard:


langkit/templates/pkg_analysis_body_ada.mako
```diff
@@ -433,6 +433,15 @@ package body ${ada_lib_name}.Analysis is
       return Implementation.Text (Unwrap_Unit (Unit));
    end Text;
```
pmderodat, line 435: Assuming we need this overload to avoid stack overflows, I think it would be better to use an unbounded string instead (more consistent with the rest of the API, especially regarding safety). What do you think?

langkit/support/langkit_support-diagnostics-output.ads
```diff
@@ -0,0 +1,37 @@
+package Langkit_Support.Diagnostics.Output is
```
pmderodat, line 1: Please add copyright headers in both files. Also, could you add a package-level docstring?

langkit/support/langkit_support-diagnostics-output.ads
```diff
@@ -0,0 +1,37 @@
...
+   --  eventually be self contained.
+
+private
+   package Offset_Vectors is new Ada.Containers.Vectors (Positive, Natural);
+
+   type Text_Buffer is record
+      Text_Ptr   : Text_Cst_Access;
+      Line_Spans : Offset_Vectors.Vector;
+      --  For each line of text in buffer, store the end line (or end of
+      --  buffer) character offset.
```
pmderodat, line 34: Please put an empty line between the two members, so that IDEs don’t consider that the docstring applies to both.

langkit/support/langkit_support-diagnostics-output.adb
```diff
@@ -0,0 +1,181 @@
...
+         --  If no next newline found, put the next offset just after the end
+         --  of the string, at the (virtual) "$" EOF character.
+         if Offset = 0 then
+            Offset := T'Last + 1;
+         end if;
+         Self.Line_Spans.Append (Offset);
+      end loop;
+
+      return  T ((if Line_Number = 1
+                 then 0
```
pmderodat, line 66: The then/else are not properly indented (one space is missing)

pmderodat: Hm… I think this is incorrect; 0 should probably be replaced with `T'First`, as IIRC text buffers could be slices, so with random boundds.

langkit/support/langkit_support-diagnostics-output.adb
```diff
@@ -0,0 +1,181 @@
...
+                 else Self.Line_Spans (Line_Number - 1)) + 1
+                 .. Self.Line_Spans (Line_Number) - 1);
```
pmderodat, line 68: I suppose you don’t want to handle this case right now, but I think the spec deserves a TODO to warn that we don’t handle CRLF line endings.

langkit/support/langkit_support-diagnostics-output.adb
```diff
@@ -0,0 +1,181 @@
...
+   --------------------------
+
+   procedure Print_Source_Listing
+     (Sloc        : Source_Location_Range;
+      Buffer      : in out Text_Buffer;
+      Lines_After : Natural := 0)
+   is
+      procedure Start_Line (Line_Nb : String := "");
+
+      Line_Nb : constant Positive := Positive (Sloc.Start_Line);
```
pmderodat, line 83: Please align the colon

langkit/support/langkit_support-diagnostics-output.adb
```diff
@@ -0,0 +1,181 @@
...
+with Ada.Text_IO; use Ada.Text_IO;
+with Ada.Wide_Wide_Text_IO; use Ada.Wide_Wide_Text_IO;
+
+with GNATCOLL.Terminal; use GNATCOLL.Terminal;
+
+with Langkit_Support.Images; use Langkit_Support.Images;
+
+package body Langkit_Support.Diagnostics.Output is
+
+   Term_Info : Terminal_Info;
```
pmderodat, line 11: Please align colons

langkit/support/langkit_support-diagnostics-output.adb
```diff
@@ -0,0 +1,181 @@
+with Ada.Strings.Wide_Wide_Fixed; use Ada.Strings.Wide_Wide_Fixed;
+with Ada.Text_IO; use Ada.Text_IO;
+with Ada.Wide_Wide_Text_IO; use Ada.Wide_Wide_Text_IO;
```
pmderodat, line 3: Please align use clauses

langkit/support/langkit_support-diagnostics-output.adb
```diff
@@ -0,0 +1,181 @@
...
+      Lines_After : Natural := 0)
+   is
+      procedure Start_Line (Line_Nb : String := "");
+
+      Line_Nb : constant Positive := Positive (Sloc.Start_Line);
+      Start_Offset : constant Positive := Positive (Sloc.Start_Column);
+      End_Offset   : constant Positive := Positive (Sloc.End_Column);
+
+      Line_Nb_Width : constant Positive :=
+        Positive'Image (Line_Nb + Lines_After)'Length - 1;
```
pmderodat, line 88: It looks like this procedure assumes that both the start and the end of `Sloc` (which should be called `Sloc_Range`, right?) are on the same line. Could you clarify if that should a precondition, or how it should be handled?

pmderodat: Ok, I just saw the comment below ;-) Could you move it to the spec?

langkit/support/langkit_support-diagnostics-output.adb
```diff
@@ -0,0 +1,181 @@
...
+      Put (Path & ":"
+           & Stripped_Image (Integer (Self.Sloc_Range.Start_Line)) & ":"
+           & Stripped_Image (Integer (Self.Sloc_Range.Start_Column)) & ":");
+      Set_Color (Term_Info, Foreground => Red);
+      Ada.Text_IO.Put (" error: ");
+      Reset_Colors;
+
+      --  Put the error message
+
+      --  TODO: Style backtick parts
```
pmderodat, line 149: Obsolete TODO?

pmderodat: Hm… could you document in the spec what this backtick feature is about?

contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -0,0 +1,114 @@
+with Ada.Directories; use Ada.Directories;
+with Ada.Exceptions;
+with Ada.Strings.Unbounded; use Ada.Strings.Unbounded;
+with Ada.Text_IO; use Ada.Text_IO;
```
pmderodat, line 4: Please align use clauses

contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -0,0 +1,114 @@
...
+
+   if Arg.Parser.Parse then
+      for File_Name of Arg.Files.Get loop
+         declare
+            Unit : constant Analysis.Analysis_Unit
+              := Ctx.Get_From_File (To_String (File_Name));
+         begin
+            if Unit.Diagnostics'Length > 0 then
+               for Diagnostic of Unit.Diagnostics loop
+                  Put_Line (To_Pretty_String (Diagnostic));
```
pmderodat, line 93: Just curious: why did not you use `Print_Diagnostic` here?

langkit/support/langkit_support-diagnostics-output.ads
raph-amiard: GS doesn't do this.

langkit/templates/pkg_analysis_body_ada.mako
raph-amiard: This is not why we have this overload, we have this overload so we can pass a pointer to the source of an unit without copying it, in order to create a `Text_Buffer` - so returning an `Unbounded_Text_Type` won't do the trick.

langkit/support/langkit_support-diagnostics-output.adb
raph-amiard: Good catch! Thanks

langkit/support/langkit_support-diagnostics-output.ads
raph-amiard: I added a comment to `Text_Ptr` anyway!

contrib/lkt/extensions/mains/lkt_toolbox.adb
raph-amiard: Good catch, that's actually a remain of the old ada test main that I did before I had `Print_Diagnostics`.

That said, I'm not sure syntax errors are going to be greatly useful with `Print_Diagnostics` because their sloc is not generally very helpful, but let's use it for consistency.

langkit/templates/pkg_analysis_body_ada.mako
raph-amiard: After live discussion: I added a warning in the comment

langkit/templates/pkg_implementation_spec_ada.mako
```diff
@@ -1268,6 +1268,18 @@ private package ${ada_lib_name}.Implementation is
...
+   --  Overload for Analysis.Text that returns an access.
+   --  
+   --  .. warning::
+   --
+   --      This is an unsafe function, that is exposed to allow efficient
+   --      access to the text buffer to Langkit_Support.Diagnostics.Output. We
+   --      plan to merge the functionality of
+   --      ``Langkit_Support.Diagnostics.Output.Text_Buffer`` that allows to
+   --      get a line of a source file into the analysis unit interface, which
+   --      will allow to get rid of this.
```
pmderodat, line 1281: Please move this documentation to the public API

langkit/support/langkit_support-diagnostics-output.adb
```diff
@@ -0,0 +1,204 @@
...
+   --------------
+   -- Get_Line --
+   --------------
+
+   function Get_Line
+     (Self        : in out Text_Buffer;
+      Line_Number : Positive) return Text_Type
+   is
+      T      : Text_Type renames Self.Text_Ptr.all;
+      Offset : Natural := 0;
```
pmderodat, line 67: I realize only now that this should be initialized to `T'First`. And thus probably that the variable should be renamed, as it’s more an index than an offset.

langkit/support/langkit_support-diagnostics-output.ads
```diff
@@ -0,0 +1,88 @@
...
+   --  Create a text buffer from a pointer to text
+
+   function Get_Line
+     (Self        : in out Text_Buffer;
+      Line_Number : Positive) return Text_Type;
+   --  Get line ``Line_Number``
+
+   procedure Print_Diagnostic
+     (Self : Diagnostic; Buffer : in out Text_Buffer; Path : String);
+   --
```
pmderodat, line 58: I suggest to remove this empty line (for consistency with the rest of the codebase)

langkit/templates/pkg_analysis_spec_ada.mako
```diff
@@ -303,6 +303,18 @@ package ${ada_lib_name}.Analysis is
    function Text (Unit : Analysis_Unit'Class) return Text_Type;
    ${ada_doc('langkit.unit_text', 3)}
 
+   function Text (Unit : Analysis_Unit'Class) return Text_Cst_Access;
+   --  Overload of Unit.Text that returns an access
```
pmderodat, line 307: Missing trailing dot here, and…

langkit/templates/pkg_implementation_spec_ada.mako
```diff
@@ -1268,6 +1268,9 @@ private package ${ada_lib_name}.Implementation is
    function Text (Unit : Internal_Unit) return Text_Type;
    --  Implementation for Analysis.Text
 
+   function Text (Unit : Internal_Unit) return Text_Cst_Access;
+   --  Overload for Analysis.Text that returns an access.
```
pmderodat, line 1272: … and extra trailing dot here ;-)

langkit/support/langkit_support-diagnostics-output.ads
```diff
@@ -0,0 +1,96 @@
...
+
+package Langkit_Support.Diagnostics.Output is
+   type Text_Buffer is private;
+   --  This type is used as a wrapper around text, where we'll keep line
+   --  information to avoid re-processing it everytime we need it.
+
+   function Create (Text : Text_Cst_Access) return Text_Buffer
+      with Pre => Text.all'First = 1;
+   --  Create a text buffer from a pointer to text. The text buffer must start
+   --  at index 0 (because we don't really care about other cases).
```
pmderodat, line 57: at index 1*

langkit/support/langkit_support-diagnostics-output.ads
```diff
@@ -0,0 +1,96 @@
...
+--      For the moment, this package has a few limitations, namely:
+--
+--      - Only source buffers with LF line endings are handled.
+--
+--      - Only error messages spanning one line are gracefully handled.
+--        eventually we want to have something more powerful that has a syntax
+--        for error spanning multiple lines.
+--
+--      - It exposes a ``Text_Buffer`` interface that we eventually want to
+--        merge in ``AnalysisUnit``. This will avoid passing an unsafe pointer
```
pmderodat, line 39: Missing underscore in AnalysisUnit

langkit/support/langkit_support-diagnostics-output.adb
```diff
@@ -0,0 +1,221 @@
...
+         if Offset = 0 then
+            Offset := T'Last + 1;
+         end if;
+         Self.Line_Spans.Append (Offset);
+      end loop;
+
+      --  Return slice from ..
+      return  T
+        (
+          --  The character just after the found newline character to ..
```
pmderodat, line 100: Mistyped ellipsis: `to...`

langkit/support/langkit_support-diagnostics-output.adb
```diff
@@ -0,0 +1,221 @@
...
+          (if Line_Number = 1
+          then 0
+          else Self.Line_Spans (Line_Number - 1)) + 1
```
pmderodat, line 103: Please add one extra space before `then` and `else` (for correct indentation)

langkit/support/langkit_support-diagnostics-output.adb
```diff
@@ -0,0 +1,221 @@
...
+   function Get_Line
+     (Self        : in out Text_Buffer;
+      Line_Number : Positive) return Text_Type
+   is
+      T      : Text_Type renames Self.Text_Ptr.all;
+
+      Offset : Natural := 0;
+      --  Offset of the currently processed line's newline char. It starts at 0
+      --  because we start the search for the next newline char from Offset + 1
+      --  in the call to `Index` below.
```
pmderodat, line 71: I still have the feeling that the code & comments would be slightly simpler if we instead tracked the index of the first character of each line (and `Offset` is really an index, not an offset), but ok, that’s much better now with the comments anyway. Thanks!

raph-amiard: In that case I invite you to try it out :) I'll rename to `Index` though, forgot this in the last pass!

PR #403, opened 2020-08-21
Preparatory work for named environments

pmderodat:


langkit/envs.py
```diff
@@ -235,25 +226,46 @@ def _parse_actions(self, actions: List[EnvAction]) -> None:
         Analyze the given list of actions and extract pre/post actions, i.e.
         actions executed before and after handling children.
         """
+        def count(cls: Type[EnvAction],
+                  sequence: Optional[List[EnvAction]] = None) -> int:
+            """
+            Return the number of ``cls`` instances in ``sequence``, or in
+            ``actions`` if ``sequence`` is left to None.
+            """
+            if sequence is None:
```
Roldak, line 235: I find this added logic a bit superfluous. Simply calling `count(..., actions)` below is way shorter and even makes it easier to understand what's going on.

langkit/envs.py
```diff
@@ -235,25 +226,46 @@ def _parse_actions(self, actions: List[EnvAction]) -> None:
...
         self.initial_env = None
         if actions and isinstance(actions[0], SetInitialEnv):
             self.initial_env = cast(SetInitialEnv, actions.pop(0))
+            first_actions.append(self.initial_env)
 
         pre, post = lsplit_by(
             lambda a: not isinstance(a, HandleChildren), actions
         )
 
+        for sequence in (pre, post):
```
Roldak, line 255: why not do the check on `actions` directly? (`count(SetInitialEnv, actions) == 0`)

langkit/templates/astnode_types_ada.mako
```diff
@@ -290,9 +292,23 @@
...
 
-   % if cls.env_spec:
+   % if cls.env_spec and cls.env_spec.actions:
 
-   <% call_prop = cls.env_spec._render_field_access %>
+   <%
+      call_prop = cls.env_spec._render_field_access
+
+      # Whether the initial env for this node is dynamic
+      has_dyn_env = cls.env_spec.initial_env
```
Roldak, line 301: `has_dyn_env = cls.env_spec.initial_env is not None` makes it clearer IMO

Roldak (changes requested):

langkit/envs.py
pmderodat: Good idea, thanks!

langkit/envs.py
pmderodat: Ah yes, you’re right. `actions` is mutated, so it looks like actually all count checks could be done directly on it. I’ll simplify this part, thanks.

langkit/templates/astnode_types_ada.mako
pmderodat: Ok, I’ll update it

Roldak (approved):

PR #406, opened 2020-08-28
Memory optimisation

raph-amiard:


langkit/support/langkit_support-vectors.adb
```diff
@@ -222,9 +226,10 @@ package body Langkit_Support.Vectors is
 
    function Pop (Self : in out Vector) return Element_Type is
       Index : constant Index_Type := Last_Index (Self);
+      Res : constant Element_Type := Get (Self, Index);
    begin
       Self.Size := Self.Size - 1;
-      return Get (Self, Index);
+      return Res;
```
pmderodat, line 232: Please add a testcase that covers both functions (see `tests/langkit_support`)

testsuite/tests/langkit_support/get_line/main.adb
```diff
@@ -14,17 +31,29 @@ procedure Main is
 
    procedure Check (Label : String; Buffer : Text_Type) is
       B           : Text_Access := new Text_Type'(Buffer);
-      TB          : Text_Buffer := Create (Text_Cst_Access (B));
+
+      TDH         : Token_Data_Handler;
```
pmderodat: Please remove extra alignment for B and TDH

testsuite/tests/langkit_support/get_line/main.adb
```diff
@@ -14,17 +31,29 @@ procedure Main is
...
+
       Lines_Count : constant Natural :=
          Ada.Strings.Wide_Wide_Fixed.Count (Buffer, (1 => Chars.LF));
+
```
pmderodat: Extra empty line

langkit/support/langkit_support-token_data_handlers.ads
```diff
@@ -160,7 +162,12 @@ package Langkit_Support.Token_Data_Handlers is
       --  token, then the second entry stands for the trivia that come after
       --  the first token, and so on.
 
-      Symbols : Symbol_Table;
+      Symbols       : Symbol_Table;
```
pmderodat, line 157: Please remove the extra alignment

langkit/support/langkit_support-token_data_handlers.ads
```diff
@@ -160,7 +162,12 @@ package Langkit_Support.Token_Data_Handlers is
...
+      --  Symbol table for this handler. Note that this can be shared accross
+      --  multiple Token_Data_Handlers.
+
+      Lines_Starts  : Index_Vectors.Vector;
```
pmderodat, line 161: Likewise

langkit/support/langkit_support-token_data_handlers.ads
```diff
@@ -160,7 +162,12 @@ package Langkit_Support.Token_Data_Handlers is
...
+      --  Table keeping count of line starts and line endings
```
pmderodat, line 162: This is not enough to understand how this data is encoded in the vector: could you add what the positive value at index `N` represents?

langkit/support/langkit_support-token_data_handlers.adb
```diff
@@ -126,7 +133,8 @@ package body Langkit_Support.Token_Data_Handlers is
               Tokens            => <>,
               Symbols           => Symbols,
               Tokens_To_Trivias => <>,
-              Trivias           => <>);
+              Trivias           => <>,
+              Lines_Starts       => <>);
```
pmderodat: Please fix the alignment

langkit/templates/pkg_analysis_spec_ada.mako
```diff
@@ -335,6 +335,10 @@ package ${ada_lib_name}.Analysis is
    procedure PP_Trivia (Unit : Analysis_Unit'Class);
    --  Debug helper: output a minimal AST with mixed trivias
 
+   function Get_Line
```
pmderodat, line 328: What about adding the `overriding` keyword to make it clear this is supposed to override the text_buffer_ifc primitive?

langkit/support/langkit_support-symbols.ads
```diff
@@ -83,6 +102,8 @@ package Langkit_Support.Symbols is
 
 private
 
+   type Thin_Symbol is new Natural;
```
pmderodat, line 89: Given that the very goal of this type is to be 32-bits wide, I think we should use Interfaces.Unsigned_32 instead of Natural

langkit/support/langkit_support-symbols.adb
```diff
@@ -75,11 +75,19 @@ package body Langkit_Support.Symbols is
...
+   ------------------------
+   -- Precomputed_Symbol --
+   ------------------------
+
+   function Precomputed_Symbol
+     (ST : Symbol_Table; Index : Precomputed_Symbol_Index) return Thin_Symbol
+   is
+   begin
       return ST.Precomputed (Index);
-      pragma Warnings (On, "value not in range");
```
pmderodat, line 82: Please restore the comment that explains why the `pragma Warnings (Off)` is needed. Also, why have you removed the `pragma Warnings (On)` part? I think it should stay.

langkit/templates/project_file.mako
```diff
@@ -181,7 +181,7 @@ library project ${lib_name} is
       Ada_Mode_Args := ();
       case Build_Mode is
          when "dev" =>
-            Ada_Mode_Args := ("-gnatwe", "-gnata");
```
pmderodat, line 184: Please restore this: we do want warnings as errors for dev setups

langkit/templates/pkg_implementation_body_ada.mako
```diff
@@ -354,12 +356,13 @@ package body ${ada_lib_name}.Implementation is
...
         (if Charset = "" then Default_Charset else Charset);
-      Symbols        : constant Symbol_Table := Create_Symbol_Table;
+      Symbols        : constant Precomputed_Symbol_Table
+        := Create_Symbol_Table;
       Context        : Internal_Context;
    begin
       Context_Pool.Acquire (Context);
       Context.Ref_Count := 1;
-      Context.Symbols := Symbols;
+      Context.Symbols := Symbol_Table (Symbols);
```
pmderodat, line 365: It looks like some users of Context.Symbols need the Precomputed_Symbol_Table while others could do with it as well, so what about changing its type, to avoid the multiple conversions here and there?

langkit/support/langkit_support-token_data_handlers.ads
```diff
@@ -113,6 +109,8 @@ package Langkit_Support.Token_Data_Handlers is
 
    package Index_Vectors is new Langkit_Support.Vectors (Positive);
 
+   type Tab_Stop_Access is access constant Positive;
```
pmderodat, line 112: When creating a TDH, the tab stop info comes from the context and is read-only: why is it materialized using an access type here, instead of a simple positive value? Also, this lacks a docstring

langkit/support/langkit_support-token_data_handlers.ads
```diff
@@ -285,4 +294,31 @@ package Langkit_Support.Token_Data_Handlers is
      (TDH : Token_Data_Handler; Line_Number : Positive) return Text_Type;
    --  Get the source text of line at index ``Line_Number``
 
+   function Get_Sloc
+     (TDH : Token_Data_Handler; Offset : Natural) return Source_Location;
+   --  Return the sloc for given ``Offset`` in ``TDH``. If:
```
pmderodat, line 299: This is not an offset, but an index of a codepoint in TDH.Source_Buffer, so I think we should rename it accordingly

langkit/support/langkit_support-token_data_handlers.ads
```diff
@@ -285,4 +294,31 @@ package Langkit_Support.Token_Data_Handlers is
...
+   --
+   --  - ``Offset`` is ``0``, return ``No_Source_Location``.
+   --
+   --  - ``Offset`` is in range ``1 .. Buffer'Last + 1``, return a
```
pmderodat, line 303: s/Buffer/TDH.Source_Buffer/ (here and below), or say above that Buffer is TDH.Source_Buffer

langkit/support/langkit_support-token_data_handlers.adb
```diff
@@ -681,16 +688,117 @@ package body Langkit_Support.Token_Data_Handlers is
...
+           Get_Line_Index (Offset, TDH.Lines_Starts);
+         Line_Offset : constant Positive := TDH.Lines_Starts.Get (Line_Index);
+      begin
+         --  Allow a sloc pointing at the EOL char (hence the + 1)
+         if Offset > TDH.Source_Buffer'Last + 1 then
+            raise Constraint_Error with "out of bound access";
+         end if;
+
+         --  Make horizontal tabulations move by stride of 8 columns, as
+         --  usually implemented in code editors.
```
pmderodat, line 744: s/8/Tab_Stop/

langkit/support/langkit_support-token_data_handlers.adb
```diff
@@ -681,16 +688,117 @@ package body Langkit_Support.Token_Data_Handlers is
...
+      is
+        (if Sought > Line_Start then After
+         elsif Sought = Line_Start then Inside
+         else Before);
+
+      function Get_Line_Index is new Floor (Positive, Index_Vectors);
+
+      Column      : Natural := 0;
+
+      Tab_Stop : Natural renames TDH.Tab_Stop.all;
```
pmderodat, line 725: s/Natural/Positive

testsuite/tests/langkit_support/get_line/main.adb
```diff
@@ -13,25 +14,32 @@ procedure Main is
...
+   procedure Check (Label : String; TDH : Token_Data_Handler);
+   function Create (Buffer : Text_Type) return Token_Data_Handler;
+
+   ------------
+   -- Create --
+   ------------
+
+   function Create (Buffer : Text_Type) return Token_Data_Handler is
+      B : Text_Access := new Text_Type'(Buffer);
+      TDH : Token_Data_Handler;
```
pmderodat, line 26: Please align colons

testsuite/tests/langkit_support/get_line/main.adb
```diff
@@ -46,17 +54,51 @@ procedure Main is
...
-      Free (B);
       New_Line;
    end Check;
 
+   --------------------------
+   -- Show_Sloc_For_Offset --
+   --------------------------
+
+   procedure Show_Sloc_For_Offset
+     (Label: String;
```
pmderodat, line 65: Please fix formatting

langkit/support/langkit_support-token_data_handlers.adb
```diff
@@ -681,16 +688,117 @@ package body Langkit_Support.Token_Data_Handlers is
...
+         Dummy_Index : Positive;
+         Line_Start  : Positive) return Relative_Position
+      is
+        (if Sought > Line_Start then After
+         elsif Sought = Line_Start then Inside
+         else Before);
+
+      function Get_Line_Index is new Floor (Positive, Index_Vectors);
+
+      Column      : Natural := 0;
```
pmderodat, line 723: Please fix colon alignments

langkit/support/langkit_support-token_data_handlers.adb
```diff
@@ -681,16 +688,117 @@ package body Langkit_Support.Token_Data_Handlers is
...
+      function Compare
+        (Sought      : Positive;
+         Dummy_Index : Positive;
+         Line_Start  : Positive) return Relative_Position
+      is
+        (if Sought > Line_Start then After
+         elsif Sought = Line_Start then Inside
+         else Before);
+
+      function Get_Line_Index is new Floor (Positive, Index_Vectors);
```
pmderodat, line 721: I find it hard to infer from its implementation what this function is supposed to compute. Please add a docstring, such as:
> Given the index of a codepoint in TDH.Source_Buffer and given TDH.Line_Starts, return the index in TDH.Source_Buffer of the first character of the line that contains this codepoint.

langkit/support/langkit_support-token_data_handlers.adb
```diff
@@ -681,16 +688,117 @@ package body Langkit_Support.Token_Data_Handlers is
...
+     (TDH   : Token_Data_Handler;
+      Token : Stored_Token_Data) return Source_Location is
+   begin
+      return Get_Sloc
+        (TDH,
+         (if Token.Source_Last < Token.Source_First
+         --  This is a special case for when the range is negative: in that
+         --  case we want to propagate that behavior to the sloc itself,
+         --  because negative ranges are used for tokens that have no
+         --  "width", and shouldn't be lookup-able, like the termination token.
```
pmderodat, line 786: Please fix the indentation of this comment block

langkit/support/langkit_support-bump_ptr.adb
```diff
@@ -135,8 +135,17 @@ package body Langkit_Support.Bump_Ptr is
...
+      -- Alloc --
+      -----------
+
+      function Alloc (Pool : Bump_Ptr_Pool) return System.Address is
+      begin
+         return
+           Allocate
+             (Pool,
+              Align (Element_T'Max_Size_In_Storage_Elements, Pointer_Size));
+      end Alloc;
```
pmderodat, line 148: This seems to duplicate the code of the other Alloc function. Could you make the other one call this one?

langkit/support/langkit_support-bump_ptr.adb
```diff
@@ -231,6 +244,18 @@ package body Langkit_Support.Bump_Ptr is
...
+        (Pool : Bump_Ptr_Pool; Length : Natural) return System.Address
+      is
+         Stride : constant Storage_Offset :=
+           Align (Element_T'Max_Size_In_Storage_Elements, Pointer_Size);
+         Size   : constant Storage_Offset := Stride * Storage_Offset (Length);
+      begin
+         return (if Length = 0
+                 then System.Null_Address
+                 else Allocate (Pool, Size));
+      end Alloc;
```
pmderodat, line 257: Likewise

langkit/support/langkit_support-diagnostics.ads
```diff
@@ -25,35 +25,98 @@ with Ada.Containers.Vectors;
...
+
+      --  IMPLEMENTATION NOTES:
+      --
+      --  The diagnostic type is tagged for backward compatibility reasons.
+      --  Diagnostic is a public API exposed langkit type, and used to be
+      --  "open", with fields corresponding to the below methods. Having
+      --  it tagged allows the client code to not change.
+      --
+      --  We want the diagnostic type to be constrained (so not have to use
+      --  Diagnostic'Class everywhere), but we have two diagnostics shape.
```
pmderodat, line 70: shapes*

langkit/support/langkit_support-diagnostics.ads
```diff
@@ -25,35 +25,98 @@ with Ada.Containers.Vectors;
...
+      --  ``Internal`` record here to store the diagnostic's data.
+
+      type Diagnostic_Kind is (Standalone, From_TDH);
+      --  There are two kinds of diagnostics: standalone diagnostics store
+      --  their slocs, whereas ``From_TDH`` diagnostics store a reference to a
+      --  ``Token_Data_Handler`` and offsets to recompute the sloc. Those are
+      --  used during lexing/parsing by langkit directly.
+
+      type Diagnostic_Internal (Kind : Diagnostic_Kind := Standalone) is record
+         Message    : Unbounded_Text_Type;
```
pmderodat, line 81: Please fix colon alignments (here and below)

contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -12,8 +12,6 @@ with Langkit_Support.Diagnostics; use Langkit_Support.Diagnostics;
 with Langkit_Support.Diagnostics.Output;
```
pmderodat, line 12: (commit message)
It’s not clear to me why this work on diagnostics is needed. Could you elaborate? Assuming this is about memory footprint: I woud have assumed that in the general case, there are mostly no diagnostics, and thus that their memory footprints was neglectible.

The biggest issue I see with this new implementation is that Diagnostic_Type instances (which exist in the public Ada API) have an unsafe reference to a TDH, yielding an unsafe API. We would need a safety net to fix this, probably cancelling the gain in memory footprint.

raph-amiard: This has been refactored to avoid this. Thanks because the result is much clearer anyway :)

langkit/templates/pkg_implementation_body_ada.mako
raph-amiard: Because most of the functions on the `Symbol_Table` are non dispatching, and are defined in such a way that a cast would be needed to call them if we stored a `Precomputed_Symbol_Table` instead. Hence we favor the common use case.

langkit/support/langkit_support-token_data_handlers.ads
raph-amiard: I think I didn't catch that the info was read-only in the context, and wanted to give a way for it to work if the tab stop was dynamically changed.

raph-amiard: Anyway, will change it!

langkit/support/langkit_support-token_data_handlers.adb
raph-amiard: You're catching typos in your own comments that I moved around :P

langkit/support/langkit_support-token_data_handlers.adb
raph-amiard: OK. Although needs to be said I copied from the other instantiations in this package, that don't have a comment and are not much clearer ... Seems OK to me because the generic function is here in the same package and documented. (Also this one has at least a descriptive name).

pmderodat: I’m pretty sure I’m the author of the other instantiation, yet no problem if think it’s not clear and ask me to document it. :-) Someone else’s eyes helping spotting missing comments and all…

IIRC, the reason I asked here is that even though I understood what the generic was supposed to do (which tells me more or less *how* Get_Line_Index is implemented, not *why*), I had I had trouble getting what was the purpose of `Get_Line_Index`, even when looking at the call site.

langkit/templates/pkg_implementation_body_ada.mako
pmderodat: Ok I see. Go ahead, thanks!

pmderodat (approved):

PR #411, opened 2020-09-09
SA28-035: Correct bug in TDH.Compute_Lines_Start

raph-amiard:


langkit/support/langkit_support-token_data_handlers.adb
```diff
@@ -170,11 +168,37 @@ package body Langkit_Support.Token_Data_Handlers is
       Idx : Natural := 0;
       --  Index in T of the newline character that ends the currently processed
       --  line.
+
+      function Index
+        (Buffer             : Text_Type;
+         Char               : Wide_Wide_Character;
+         Start_Pos, End_Pos : Positive) return Natural;
+      --  Helper function. Return first occurence of ``Char`` in ``Buffer
```
pmderodat, line 176: Return the index* of the first occurence […]. I think this should also mention that it returns 0 if there is no such occurence.

BTW, what about instead changing the declaration of `T` above so that it covers only the part of `Source_Buffer` that is initialized, and remove the `End_Pos` argument here?

raph-amiard: > BTW, what about instead changing the declaration of T above so that it covers only the part of Source_Buffer that is initialized, and remove the End_Pos argument here?


I don't see the point. I must pass Start_Pos anyway.

pmderodat (dismissed):

langkit/support/langkit_support-token_data_handlers.adb
```diff
@@ -170,11 +168,37 @@ package body Langkit_Support.Token_Data_Handlers is
       Idx : Natural := 0;
       --  Index in T of the newline character that ends the currently processed
       --  line.
+
+      function Index
+        (Buffer             : Text_Type;
+         Char               : Wide_Wide_Character;
+         Start_Pos, End_Pos : Positive) return Natural;
```
pmderodat, line 175: Given the errors in the travis, I guess you need to make `End_Pos` a Natural (not a positive) for empty buffers.

pmderodat (approved):

PR #412, opened 2020-09-09
manage.py: restore mistakenly removed "-dtmp" argument

pmderodat:


raph-amiard:
It wasn't mistakenly removed. I removed it because I don't think it's a great default. I want to choose when I want to keep all the test temporaries. Often I just don't because they take up a lot of space. Ready to discuss this though :)

pmderodat:
It was mistakenly removed in an unrelated commit, then. :-) (in commit dc3b3fdb86ca1fb8383c7904781cf00e19caa099
 “SA28-035: symbols on 32bits”) Knowing you did that would have saved me 10min wondering why we no longer did that automatically and worrying that I broke e3-testsuite that way…

Anyway, disk space for this was never an issue for me: only now I had a look at the space for a full testsuite run, which is indeed huge (currently ~3GB). So even though always preserving the temporary files is very convenient for me, I agree it’s not a good default: I’ll just workaround this on my side. Thanks.

PR #413, opened 2020-09-10
Allow passing args to python App programmatically

raph-amiard:


pmderodat (approved):

pmderodat (changes requested):
There is a minor style check issue. LGTM after that, thanks!

pmderodat (approved):

PR #415, opened 2020-09-14
Fix printing of Tokens in gdb helpers.

Roldak:
Due to recent changes in the representation of tokens in Langkit,
the sloc_range of a token is not a field anymore and must now be
computed whenever it needs to be retrieved.


langkit/gdb/tdh.py
```diff
@@ -65,7 +65,10 @@ def kind(self):
 
     @property
     def sloc_range(self):
-        return SlocRange(self.value['sloc_range'])
+        sloc_range_sym = "langkit_support__token_data_handlers__sloc_range"
+        return SlocRange(gdb.lookup_global_symbol(sloc_range_sym).value()(
+            self.tdh.value, self.value
+        ))
```
pmderodat, line 71: Pretty-printers should not try to execute inferior code, both for performance and to avoid semi-automatic process corruption when data structures are inconsistent. I’m afraid we need to re-implement the Sloc_Range method here.

langkit/gdb/tdh.py
```diff
@@ -47,6 +51,42 @@ def trivia(self, token_no, trivia_no):
...
+        line_offset = int(self._line_starts[line])
+
+        # Get the last character index of the buffer
+        source_last = int(self.value['source_last'])
+
+        # Compute the column number as being the given character index minus
+        # the character index of the first character of the line. Make sure
+        # the character index is not higher than the index of the last
+        # character of the buffer.
+        column = min(char_index, source_last) - line_offset
```
pmderodat, line 85: I think it’s ok to have this limitation, but please add a TODO that explains that we don’t support tab expansion here

pmderodat (changes requested):
Thanks a lot for fixing this! I think this change is OK with a couple of minor improvements.

pmderodat (approved):

PR #422, opened 2020-10-01
$.Introspection: introduce generic operations on arrays

pmderodat:


langkit/documentation.py
```diff
@@ -248,6 +248,14 @@ def instantiate_templates(doc_dict: Dict[str, str]) -> DocDatabase:
         Raised when introspection functions (``${ctx.lib_name}.Introspection``)
         are improperly used to evaluate a node data (field or property).
     """,
+    'langkit.introspection.bad_type_error': """
+        Raised when introspection functions (``${ctx.lib_name}.Introspection``)
+        are given values with inconsistent types.
+    """,
+    'langkit.introspection.out_of_bounds_error': """
+        Raised when introspection functions (``${ctx.lib_name}.Introspection``)
+        are queried an out-of-bound index access on an array.
```
raph-amiard, line 257: 1. I'm not sure of the value of having exceptions with so specific purposes, what use case do you see against just one for introspection errors?
2. The formulation of this message is weird: "are passed an index that generates an out of bound access to an array".

EDIT: About 1., see comment above about type errors

langkit/documentation.py
```diff
@@ -248,6 +248,14 @@ def instantiate_templates(doc_dict: Dict[str, str]) -> DocDatabase:
         Raised when introspection functions (``${ctx.lib_name}.Introspection``)
         are improperly used to evaluate a node data (field or property).
     """,
+    'langkit.introspection.bad_type_error': """
```
raph-amiard, line 251: The fact of having a `Bad_Type_Error` doesn't seem very consistent with the fact that:

1. Access to fields on the wrong type will raise a `Node_Data_Evaluation_Error`:

```
         (raise Node_Data_Evaluation_Error with "no such field on this node");
```

2. Querying the `Index` of a field on the wrong node type will raise a `Constraint_Error`

3. The fact that type constraints are called `Value_Constraint`s.

TBH I would much rather type constraints are called `Type_Constraint` and the two case above raise a `Bad_Type_Error`, that would be much easier to understand and deal with, the current design is inconsistent IMO.

EDIT: Specifically for the case of `Constraint_Error`, every error here on `Value_Type` could be a constraint error, but we'd rather not do that for general design reasons, so I think we should just not use it.

langkit/templates/pkg_introspection_body_ada.mako
```diff
@@ -541,6 +541,131 @@ package body ${ada_lib_name}.Introspection is
...
+   begin
+      case Kind (Self) is
+         % for t in array_types:
+            when ${t.introspection_kind} =>
+               declare
+                  A : ${t.api_name} renames
+                     Self.Value.Value.${t.introspection_kind}.all;
+               begin
+                  if Index not in A'Range then
+                     raise Out_Of_Bounds_Error with "index of of array bounds";
```
raph-amiard, line 600: typo -> "index of of array bounds"

langkit/templates/pkg_introspection_spec_ada.mako
```diff
@@ -145,6 +145,35 @@ package ${ada_lib_name}.Introspection is
...
+
+   function Array_Element_Constraint
+     (Kind : Array_Value_Kind) return Value_Constraint;
+   --  Return the constraint for elements of Kind arrays
+
+   function Array_Length (Self : Value_Type) return Natural;
+   --  Assuming that Self is an array (regardless of its element type), return
+   --  the number of elements it contains.
+   --
+   --  This raises a Bad_Type_Error if Value is not an array.
```
raph-amiard, line 160: Please use sphinx syntax for references to code entities: 
```
``Bad_Type_Error``
```

raph-amiard (changes requested):

langkit/documentation.py
pmderodat: Thinking more about it, the description is too specific. I’ll try “are passed an out of bounds index”. As discussed live, we’ll keep this exception, but merge most others into the “bad type” one.

langkit/documentation.py
pmderodat: As discussed live: I’ll merge `Node_Data_Evaluation_Error` and `Invalid_Field` into `Bad_Type_Error`. In addition, I’ll replace most `Constraint_Error` with `Bad_Type_Error` too. And let’s go for the `Value_Constraint` renaming.

raph-amiard (approved):

PR #442, opened 2020-11-09
Fix owership share leak when memoization is cancelled after cache reset

pmderodat:


langkit/templates/memoization_ada.mako
```diff
@@ -302,19 +304,19 @@ end Find_Memoized_Value;
...
+   Value  : Mmz_Value;
+   Stored : out Boolean) is
 begin
    --  If Handle was created using a memoization map that has been since then
    --  reset, do nothing: the result can be partly stale due to the event that
    --  triggered the memoization tables reset.
 
-   if Handle.Cache_Version < Unit.Cache_Version then
-      return;
+   Stored := Unit.Cache_Version <= Handle.Cache_Version;
```
Roldak, line 316: This is not the inverse of the original condition (`<=` instead of `>=`). Not sure if it's expected or not, just mentioning in case it's a typo.

Roldak (approved):

langkit/templates/memoization_ada.mako
pmderodat: FTR, the operands are switched, so this *is* the negation of the original condition. Sorry for the confusion, and thanks for the review!. ;-)

PR #444, opened 2020-11-25
Add pass to generate railroad diagrams

raph-amiard:


langkit/compile_context.py
```diff
@@ -1702,23 +1717,36 @@ def emit(self, lib_root, check_only=False, warnings=None,
...
+            self.all_passes.extend(self.code_emission_passes())
 
             # Run plugin passes at the end of the pipeline
-            all_passes.extend(plugin_passes)
+            self.all_passes.extend(plugin_passes)
+
+        for p in self.all_passes:
+            if p.is_optional and p.name in explicit_passes_triggers.keys():
+                trig = explicit_passes_triggers[p.name]
+                printcol(
```
pmderodat, line 1722: Could we enable this printing only when verbosity level is “debug”?

REQUIREMENTS.dev
```diff
@@ -11,3 +11,5 @@ yapf==0.30.0
 pycodestyle
 mypy==0.780
 flake8
+railroad-diagrams~=1.1.0
+
```
pmderodat, line 15: Extra empty line

langkit/passes.py
```diff
@@ -92,9 +92,31 @@ class AbstractPass:
...
+    """
+
+    doc: str
+
     def __init__(self, name: str, disabled: bool = False) -> None:
         self.name = name
         self.disabled = disabled
+        self.is_optional = False
+
+    def optional(self, doc: str, disabled: bool = True) -> AbstractPass:
```
pmderodat, line 108: Given that this does not create a new pass (which we would not need anyway), what about just adding doc/optional arguments to the constructor? Then for instance have `disabled: Optional[bool] = None`, which results in `disabled = True` if optional and `disabled = False` otherwise.

langkit/compile_context.py
```diff
@@ -2044,6 +2045,12 @@ def code_emission_passes(self):
             EmitterPass('instrument for code coverage',
                         Emitter.instrument_for_coverage),
 
+            GrammarRulePass('emit railroad diagrams', emit_railroad_diagram)
+            .optional("""
+            Emit SVG railroad diagrams for grammar rules, in share/doc. Need
```
pmderodat, line 2043: Needs*

langkit/compile_context.py
```diff
@@ -2044,6 +2045,12 @@ def code_emission_passes(self):
...
+            the railroad-diagrams python library.
```
pmderodat, line 2044: Pythohn*

langkit/railroad_diagrams.py
```diff
@@ -0,0 +1,98 @@
...
+
+    Railroads will be emitted in $BUILD/doc/svg.
+    """
+    try:
+        from railroad import (
+            Choice, Diagram, DiagramItem, End, OneOrMore, Optional,
+            Sequence, Skip, Start, ZeroOrMore
+        )
+    except ImportError:
+        return
```
pmderodat, line 22: This pass is disabled by default, and I think it would be good to have it actually fail if requested while railroad is missing, so I would suggest remove the try/except/return guard.

langkit/railroad_diagrams.py
```diff
@@ -0,0 +1,98 @@
...
+
+    d = Diagram(
+        # Explicit start point with the parser's name as label
+        Start("simple", label=parser.name.lower()),
+        *[c for c in [recurse(parser)] if c is not None],
+        End("simple")
+    )
+
+    # Output the diagram to svg in $BUILD/doc/svg/$RULENAME.svg
+    out_dir = P.join(get_context().emitter.lib_root, "doc", "svg")
```
pmderodat, line 94: Please add a subdirectory that corresponds to the generated library (using `emitter.lib_name_low`)

Also, since `pathlib.Path` is involved, this block could be greatly simplified:

```python
out_dir = pathlib.Path(…, "doc", "svg")
out_dir.mkdir(parents=True, exist_ok=True)
with (out_dir / …).open("w"):
    …
```

langkit/railroad_diagrams.py
```diff
@@ -0,0 +1,98 @@
...
+from langkit.parsers import (
+    Defer, List, Opt, Or, Parser, _Extract, _Row, _Token, _Transform
+)
+
+
+def emit_railroad_diagram(parser: Parser) -> None:
+    """
+    Pass to emit railroad diagrams using the railroad_diagrams library.
+
+    Railroads will be emitted in $BUILD/doc/svg.
```
pmderodat, line 14: Output path needs an update

langkit/libmanage.py
```diff
@@ -26,62 +29,92 @@
...
+    """
+
     # The create_context method will create the context and set it
     # only right before executing commands.
     context: 'CompileCtx'
     """
     Langkit compilation context.
     """
 
+    # This will be set in the run method, when we have parsed arguments
```
pmderodat, line 114: `run` method
(IIRC we have agreed to quote references to code)

langkit/libmanage.py
```diff
@@ -97,6 +130,8 @@ def __init__(self, root_dir: Optional[str] = None) -> None:
             )
         )
 
+        self.build_mode = ""
+
```
pmderodat, line 134: This is not a valid build mode, why is this needed?

langkit/libmanage.py
```diff
@@ -1159,7 +1191,7 @@ def do_list_optional_passes(self, args):
                 print(p.doc)
                 print()
 
-    def do_help(self, args):
+    def do_help(self, args: argparse.Namespace) -> None:
         """
         Print usage and exit.
```
pmderodat, line 1192: Type annotation should be removed below

langkit/libmanage.py
```diff
@@ -1195,51 +1231,52 @@ def setup_environment(self, build_mode, add_path):
...
-        :param str name: Name of the process to run, use for error message
+        :param name: Name of the process to run, use for error message
             formatting only.
-        :param list[str] argv: Arguments for the command to run.
-        :param dict[str, str]|None env: Environment to use for the command to
-            run. If None, use self.derived_env().
+        :param argv: Arguments for the command to run.
+        :param env: Environment to use for the command to run. If None, use
+            self.derived_env().
         :param bool abort_on_error: If the command stops with an error, exit
```
pmderodat, line 1269: Type annotation left

langkit/compile_context.py
```diff
@@ -2052,8 +2045,14 @@ def code_emission_passes(self):
             """),
 
             GlobalPass('report unused documentation entries',
-                       lambda ctx: ctx.documentations.report_unused(),
-                       disabled=not self.report_unused_documentation_entries),
+                       lambda ctx: ctx.documentations.report_unused())
+            .optional(
+                """
+                Report unused documentation entries. This is an internal pass
+                that is used by langkit devs to maintain the langkit
```
pmderodat, line 2052: Langkit*

langkit/libmanage.py
```diff
@@ -71,6 +75,13 @@ class ManageScript:
     Whether warnings to build the generated library are enabled by default.
     """
 
+    # The create_context method will create the context and set it
+    # only right before executing commands.
+    context: 'CompileCtx'
```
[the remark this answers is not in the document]
raph-amiard: Aaah thanks! that's why I had to use a string ref.

langkit/compile_context.py
```diff
@@ -1691,6 +1703,9 @@ def emit(self, lib_root, check_only=False, warnings=None,
         self.report_unused_documentation_entries = (
             report_unused_documentation_entries
         )
+        self.check_only = check_only
+
+        explicit_passes_triggers = kwargs.pop('explicit_passes_triggers', {})
```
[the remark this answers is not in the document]
raph-amiard: I must admit I mostly copied the behavior of the surrounding code: how do you determine whether it's necessary in this case to pass via kwargs? I must be missing something.

langkit/compile_context.py
```diff
@@ -1636,11 +1648,11 @@ def load_plugin_pass(pass_or_name):
         assert isinstance(result, AbstractPass)
         return result
 
-    def emit(self, lib_root, check_only=False, warnings=None,
-             report_unused_documentation_entries=False,
-             default_max_call_depth=1000, **kwargs):
+    def prepare(self, lib_root, check_only=False, warnings=None,
```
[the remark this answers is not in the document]
raph-amiard: SGTM, done

langkit/compile_context.py
raph-amiard: I don't know, if you don't add that you can pass anything to `--pass-on/off` and you won't know if it's working or not. Another solution would be to emit an error message if the user passes a wrong pass name, but I prefered the info message solution

langkit/passes.py
raph-amiard: I didn't do that because it requires much more boilerplate and refactoring:

* Adding the `optional` argument to every constructor that requires it (so potentially every subclass)
* Forwarding the arguments to the super call

For not much practical benefits

langkit/compile_context.py
pmderodat: I think we should definitely stop with an error if there is a typo in the name of the pass (just like we must stop with an error if there is a typo in an option name rather than ignore it with a warning). But then indeed no reason to print the enabling status for all optional passes.

langkit/passes.py
pmderodat: Ok. I still prefer the constructor arg (as I did for `disabled`), as I find the uses of `.optional` awkward, but I have no problem considering this as a matter of taste, so I won’t object.

langkit/railroad_diagrams.py
raph-amiard: > Please add a subdirectory that corresponds to the generated library

Why ? I don't understand, it was my impression that we were *not* trying to mimic install trees anymore in `build/` ?

langkit/passes.py
pmderodat: I forgot to add: regarding docstring enforcement/runtime check: I guess we could enforce doc for optional passes in constructor with Mypy’s overloads, but that does not change my conclusion (I won’t object).

langkit/libmanage.py
raph-amiard: We have agreed to quote references to code in docstrings, where this will actually have an impact on generated documentation. I don't think we ever agreed to do that in comments, where it doesn't make a whole lot of sense IMO. I'm OK to discuss it but that's a different issue in my view.

langkit/libmanage.py
raph-amiard: It's not, I think it's a leftover, I'll remove it.

langkit/railroad_diagrams.py
pmderodat: It’s true we don’t try anymore for anything that’s under GPRbuild/GPRinstall’s control (sources, build artifacts), but the install procedure is manual for everything else. Given that we just merge `share` into the prefix at install type (in our manage scripts), `share` must already have the correct layout. The other option is to generate the SVG outside of this directory and amend the `do_install` function install it in an appropriate place.

langkit/libmanage.py
pmderodat: Ok, let’s discuss :) I did not realize we agreed to do that only for doc generation, and I’m fine with this refinment. However in other places, I would advise to surround with double quotes or backquotes as soon as it’s not obvious that the name is a code entity (mostly: when there’s no underscore in the name, so apply to `run` here), as it helps reading (at least to me). WDYT?

langkit/railroad_diagrams.py
raph-amiard: Ok but to be honest, for the moment those are just build artifacts, and meant to be then leveraged in doc generation, which isn't even part of the build tree. I don't think we want to ship those as is. WDYT ?

pmderodat: Everything in `$BUILD/share` is installed, so if this is just a build artifact, maybe generate it in `$BUILD/obj`?

langkit/compile_context.py
raph-amiard: OK, I'll do that

langkit/passes.py
raph-amiard: Ok

langkit/railroad_diagrams.py
raph-amiard: Yeah but it's more like source, ie. I don't want it to disappear on gprclean, since it's generated by `manage.py`

langkit/libmanage.py
raph-amiard: I'm fine with it I guess, which syntax do you want to use ?

langkit/railroad_diagrams.py
pmderodat: gprclean does not removes all files in object directories, and `$BUILD/obj/` isn’t the actual object directory anyway (it’s `$BUILD/obj/$BUILD_MODE/` anyway, so there should be no problem. If it’s a matter of principle, then just generate the SVG directly in `$BUILD/`?

langkit/libmanage.py
pmderodat: On one hand I find double quotes easier to type, but on the other hand it’s inconsistent with the backquotes we use everywhere else (i.e. in docstrings for for Markdown). I’m fine with both options as long as we stick to the rule. Do you have a preference, in the light of this?

langkit/railroad_diagrams.py
raph-amiard: But it is in `$BUILD` ... (in `$BUILD/doc` but not in `share` anyway, so I still don't really see the problem)

pmderodat: Ah, my bad then… there is a `doc` dir in `share`, so I suppose I confused both. What about `$BUILD/parser-railroads/$RULE.svg` to avoid the confusion?

langkit/libmanage.py
raph-amiard: I dunno :( Also the doc style is *double* backquotes, which is even more tedious to type

langkit/railroad_diagrams.py
raph-amiard: Ok

langkit/libmanage.py
raph-amiard: Let's go for ""

langkit/compile_context.py
```diff
@@ -1703,23 +1719,34 @@ def emit(self, lib_root, check_only=False, warnings=None,
...
 
-            all_passes.extend(self.code_emission_passes(annotate_fields_types))
+            self.all_passes.extend(self.code_emission_passes())
 
             # Run plugin passes at the end of the pipeline
-            all_passes.extend(plugin_passes)
+            self.all_passes.extend(plugin_passes)
+
+        for p in self.all_passes:
+            if p.is_optional and p.name in explicit_passes_triggers.keys():
```
pmderodat, line 1722: I did not realize this until I saw the testcase: having to pass `--pass-on "foo bar"` (i.e. having spaces in pass names) will be cumbersome to use. What about introducing the concept of “command-line name” in pass classes, for instance replacing spaces with dashes and converting to lowercase?

raph-amiard: I'm OK to do this in a subsequent PR

pmderodat (approved):

PR #453, opened 2021-01-07
Make it possible to add an env entry to a named environment 

pmderodat:


langkit/envs.py
```diff
@@ -166,6 +172,31 @@ def add_to_env_kv(key: AbstractExpression,
     )
 
 
+def add_to_env_by_name(key: AbstractExpression,
+                       val: AbstractExpression,
+                       name_expr: AbstractExpression,
```
Roldak, line 177: Could we remove the `_expr` part from these two parameter names? It feels weird for the users when calling this function using keyword argument notation.

langkit/templates/pkg_implementation_body_ada.mako
```diff
@@ -1336,60 +1364,56 @@ package body ${ada_lib_name}.Implementation is
...
+      elsif
+         --  If requested, reject foreign destination environments.  Note that
+         --  this checks only explicit destination environments
+         --  (Dest_Env_Fallback): Set_Initial_Env already sanitized initial
+         --  environments (State.Current_Env).
+         --
+         --  This is an attempt at identifying uses of the unsound relocation
+         --  mechanism (as opposed to named environments), so this applies to
+         --  all foreign environments (root scope included).
+         DSL_Location'Length > 0
```
Roldak, line 1382: I would be in favor of adding another condition: `and then Dest_NED = null`, so that this check is only done if we actually did go through the fallback. This allows reusing the old code (before the named envs) as the fallback code to handle the non-named cases.

Roldak (changes requested):

langkit/envs.py
pmderodat: We already have such suffixes in other functions from `langkit.envs`. I’m ok with changing them too, but that’s going to be in another commit.

langkit/templates/pkg_implementation_body_ada.mako
pmderodat: As discussed live: when there is an env name, we don’t evaluate the fallback env, so `Dest_Env_Fallback` is `Empty_Env`, and thus the check is not triggered in that case. So the extra condition you suggest is redundant. I’ve added a comment to explicitly state this. Thanks!

Roldak (approved):

PR #459, opened 2021-01-13
$.Analysis: turn Get_From_*/Get_With_Error preconds into regular code

pmderodat:


raph-amiard (commented):
That's a bit of a weird pattern. Is the thing that we'd like those pre enabled in production builds ? If that's the case, isn't there a way to do just that via `Assertion_Policy` ?

pmderodat:
IIRC, when we discussed this, you also had the argument that contracts are not debug friendly: whole expression evaluates to false => you just get an assertion_failure exception, whereas with explicit checks, you can use a custom exception message.

pmderodat:
Another point in favor of explicit check is that we can share them with other APIs. To achieve this, here I should move the checks to $.Implementation (no point in having contracts here as users won’t see them). What do you think?

raph-amiard:
I agree with everything, especially the point about factoring that in the implem.

I was just thinking about the pattern:

```ada
Pre => Cond or else raise Assertion_Error with "..."
```

That allows to have better diagnostics in pre conditions, but here, factoring in the implem seems even better, especially with the added constraint of prod vs debug.

raph-amiard (approved):

PR #465, opened 2021-01-20
Work on case insensitivity

pmderodat:


support/langkit_support-text.adb
```diff
@@ -201,4 +202,26 @@ package body Langkit_Support.Text is
       return Decode (S, "UTF-8");
    end From_UTF8;
 
+   --------------
```
raph-amiard, line 205: It's really non obvious to me where you use that. I imagine in the last commit ? 

In that case this again seems like artificial splitting of your work: It's hard to understand the context when reviewing.

raph-amiard (approved):

support/langkit_support-text.adb
pmderodat: It’s a code import from LAL, thus first used in LAL, and then indeed will be reused in the last commit. I’ll amend the commit message, I forgot to complete it.

raph-amiard (approved):

PR #474, opened 2021-02-04
To_Unbounded_Text (S: String)

leogermond:
Simple overloading of `To_Unbounded_Text` to accept a `String` argument. The implementation is very straightforwards but may be a bit lackluster performance-wise.

By overloading the `To_Unbounded_Text` to accept `String`, calls to` To_Unbounded_String ( F ) `can become ambiguous, if `F` is an overloaded function that can return both `String` and `Text_Type` ( `Wide_Wide_String`).
The solution in that case is to use the `Text_Type` qualifier, as the `Text_Type` version is more efficient (it is already Unicode).

### DIFF
--- a/langkit/compiled_types.py
+++ b/langkit/compiled_types.py
@@ -4234,7 +4234,7 @@ class SymbolType(CompiledType):
         ])
 
     def to_public_expr(self, internal_expr):
-        return 'To_Unbounded_Text (Image ({}))'.format(internal_expr)
+        return "To_Unbounded_Text (Text_Type'(Image ({})))".format(internal_expr)
 
     def to_internal_expr(self, public_expr, context):
         return 'Lookup_Symbol ({}, To_Text ({}))'.format(context, public_expr)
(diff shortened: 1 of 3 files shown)


pmderodat:
Hello Léo, thanks for your contribution.

The CI detected regressions in the testsuite: could you take care of fixing them? Also, it seems that the first commit breaks things that the second commit fixes: we would like to keep Langkit working at each commit as much as possible, so could you merge the two commits? Thanks!

langkit/compiled_types.py
```diff
@@ -4234,7 +4234,7 @@ def __init__(self):
         ])
 
     def to_public_expr(self, internal_expr):
-        return 'To_Unbounded_Text (Image ({}))'.format(internal_expr)
+        return 'To_Unbounded_Text (Text_Type\'(Image ({})))'.format(internal_expr)
```
pmderodat, line 4237: Can you replace the single quotes that are string delimiters into double quotes, to avoid the need for escaping?

leogermond:
There is issues with string literals actually, so I'm putting this in draft for the moment.

leogermond:
That is too much of a hassle to handle literals with that change

PR #496, opened 2021-04-19
Start implementing a generic API for all Langkit-generated libraries

pmderodat:


support/langkit_support-generic_api.ads
```diff
@@ -0,0 +1,76 @@
...
+
+   type Language_Id is private;
+   --  Unique identifier for a Langkit-generated library
+
+   function Language_Name (Id : Language_Id) return String;
+   --  Return the name of the language that the library corresponding to ``Id``
+   --  analyzes.
+
+   type Any_Grammar_Rule is new Natural;
+   subtype Grammar_Rule is Any_Grammar_Rule range 1 .. Any_Grammar_Rule'Last;
```
raph-amiard, line 42: What's the point of having it be a public `Natural` rather than a private type ?

support/langkit_support-generic_api.ads
```diff
@@ -0,0 +1,76 @@
...
+   function Last_Grammar_Rule (Id : Language_Id) return Grammar_Rule;
+   --  Return the last grammar rule for the given language
+
+   function Default_Grammar_Rule (Id : Language_Id) return Grammar_Rule;
+   --  Return the default grammar rule for the given language
+
+   function Grammar_Rule_Name
+     (Id : Language_Id; Rule : Grammar_Rule) return String;
+   --  Return the camel-with-underscores name for the given grammar rule
+   --  according to the given language. Raise a Precondition_Failure exception
```
raph-amiard, line 64: Missing `` around the exception name. Generally missing those everywhere here :)

support/langkit_support-generic_api.ads
raph-amiard: More generally I wonder if it wouldn't be better to have those grammar rules encapsulate the language id. The fact that you can (and will) have a `Grammar_Rule` instance that is valid in two different languages but means something different is a bit unsettling to me.

support/langkit_support-generic_api-analysis.ads
```diff
@@ -0,0 +1,270 @@
...
+      return Analysis_Context;
+   --  Create a new analysis context.
+   --
+   --  ``Charset`` will be used as a default charset to decode input sources
+   --  in analysis units. Please see ``GNATCOLL.Iconv`` for several supported
+   --  charsets. Be careful: passing an unsupported charset is not guaranteed
+   --  to raise an error here. If no charset is provided, use the language
+   --  default.
+   --
+   --  When ``With_Trivia`` is true, the parsed analysis units will contain
```
raph-amiard, line 78: We're probably missing a definition of trivia in user doc. Can you put at least a TODO here ? I'll see if there are other language agnostic concepts that we don't explain

support/langkit_support-generic_api-analysis.ads
```diff
@@ -0,0 +1,270 @@
...
+   --  Return the root node for this unit, or ``No_Analysis_Node`` if there is
+   --  none.
+
+   --  TODO??? Bind all other analysis unit primitives
+
+   -------------------------------
+   -- Analysis nodes operations --
+   -------------------------------
+
+   function Is_Null (Node : Analysis_Node'Class) return Boolean;
```
raph-amiard, line 142: Not fond of the name `Analysis_Node`, it's verbose for no good reason: `Analysis` doesn't provide useful information about what it is.

What about just `Node` ?

support/langkit_support-generic_api-analysis.ads
```diff
@@ -0,0 +1,270 @@
...
+   --  Return whether ``Node`` is a null node reference
+
+   function "=" (Left, Right : Analysis_Node'Class) return Boolean;
+   --  Return whether ``Left`` and ``Right`` designate the same node
+
+   function Image (Node : Analysis_Node'Class) return String;
+   --  Return a short string describing ``Node``, or None" if ``Node.Is_Null``
```
raph-amiard, line 149: `None"` is probably ` ``None`` `

raph-amiard: Also I'd rather you call ``Node`` ``Self``, to be consistent amongst different types in this API.

support/langkit_support-generic_api-analysis.ads
```diff
@@ -0,0 +1,270 @@
...
+
+   ------------------------------
+   -- Analysis unit operations --
+   ------------------------------
+
+   overriding function Get_Line
+     (Self : Analysis_Unit; Line_Number : Positive) return Text_Type;
+   --  Return the line of text at line number ``Line_Number``
+
+   function Root (Unit : Analysis_Unit'Class) return Analysis_Node;
```
raph-amiard, line 132: I think having a link back to the Unit would be good here, I could see that as being generally useful in the generic API.

support/langkit_support-generic_api.ads
pmderodat: As discussed live: having a scalar allows super easy iteration & debugging of invalid values for users. It’s also possible with a private type that encapsulates the language id + the scalar, but then we need to check the language id everywhere we accept a grammar rule + the extra “rule_to_language” accessor to ease debugging on users’ side. We agreed to keep the current type and rename it `Grammar_Rule_Index`.

support/langkit_support-generic_api-analysis.ads
pmderodat: Trivias are defined in LAL’s manual (https://docs.adacore.com/live/wave/libadalang/html/libadalang_ug/core_concepts.html#tokens-and-trivias) but also in Langkit_Support (https://github.com/AdaCore/langkit/blob/master/support/langkit_support-token_data_handlers.ads#L53). What would you suggest in addition?

support/langkit_support-generic_api-analysis.ads
pmderodat: As discussed live: it’s too likely that `Node` will conflict as a type name with variable names, so any prefix has value, even though it does not need to be `Analysis_` (had the advantage of being consistent with contexts and units). As discussed live, let’s rename all prefixes to `Lk_`.

support/langkit_support-generic_api-analysis.ads
raph-amiard: I'm not sure what I suggest TBH, but clearly the current situation is not great from a langkit perspective. We're lacking an explanation of some general langkit concepts, and if you try to use langkit libraries generically, you might not think of going into LAL's user guide, nor in token_data_handlers.ads line 53

support/langkit_support-generic_api-analysis.ads
pmderodat: Sure, I’ll also add the unit -> context backlink

support/langkit_support-generic_api-analysis.ads
pmderodat: Ok, I’ll add a general TODO about this at the top of this unit

raph-amiard (approved):

PR #513, opened 2021-06-21
Lexical envs: cache empty envs in dynamic env getters

pmderodat:


support/langkit_support-lexical_envs_impl.adb
```diff
@@ -257,8 +257,14 @@ package body Langkit_Support.Lexical_Envs_Impl is
          --  possible.
          if Cache_Enabled and then Self.Env /= Null_Lexical_Env then
 
-            --  If it is not stale, return it
-            if not Is_Stale (Self.Env) then
+            --  If it is not stale, return it. Take care of the special case of
+            --  Empty_Env, which are always stale, but we try to cache them
```
raph-amiard, line 261: I'm not sure about this comment: How can you say they're always stale since they're clearly not (they're not if the context version has not been bumped) ?

Or you're just talking about the result of the `Is_Stale` function, but that's still pretty unclear IMO.

raph-amiard: What's annoying here is that:

1. `Is_Stale` exists *solely* for the purpose of this `Get_Env` function, and to determine this logic.
2. Here, you have to put a bit of this logic out of `Is_Stale` because of the post-processing you're doing on the Empty_Env.

I think that's part of the reason I wanted to get rid of `Empty_Env` originally, so that you can keep the staleness logic in `Is_Stale`, and you don't try to unify `Empty_Env` anymore, you just have an `Is_Empty` predicate, so you can just return the cached env here.

pmderodat: As discussed live: I did not realize that `Is_Stale` was only used in `Get_Env`, thanks! I’ll move `Is_Stale` to be nested in `Get_Env` so that I can move that logic there.

I agree that not having to unify `Empty_Env` would be even nicer on the principle, but it’s unclear if it’s doable without an earthquake: we would need to make the null expr for LexicalEnv in the DSL create different values for EmptyEnv through time (different versions) and then we would have tons of places where we have potentially obsolete empty environments in envs, other caches, etc.

Probably something to tackle later on…

raph-amiard (approved):

PR #518, opened 2021-07-12
U712-008: Add Event_Handler

raph-amiard:


langkit/documentation.py
```diff
@@ -686,6 +686,45 @@ def instantiate_templates(doc_dict: Dict[str, str]) -> DocDatabase:
         buffer for a requested filename.
     """,
 
+    #
+    # Event handlers
+    #
+    'langkit.create_event_handler': """
+        Create a event handler. When done with it, the result must be passed
```
pmderodat, line 693: an* event (I noticed other occurences in the patch)

langkit/documentation.py
```diff
@@ -686,6 +686,45 @@ def instantiate_templates(doc_dict: Dict[str, str]) -> DocDatabase:
...
+        to ``${capi.get_name('dec_ref_event_handler')}``.
+
+        Pass as ``data`` a pointer to hold your private data: it will be passed
+        to all callbacks below.
+
+        ``destroy`` is a callback that is called by
+        ``${capi.get_name('dec_ref_event_handler')}`` to leave a chance to
+        free resources that ``data`` may hold.
+
+        TODO add doc on callbacks
```
pmderodat, line 703: Forgotten bit? Documenting callback arguments would indeed be great.

langkit/documentation.py
```diff
@@ -686,6 +686,45 @@ def instantiate_templates(doc_dict: Dict[str, str]) -> DocDatabase:
...
+    """,
+    'langkit.event_handler_type': """
+        Interface to handle events sent by the AnalysisContext
```
pmderodat, line 706: analysis context*

langkit/templates/c_api/header_c.mako
```diff
@@ -183,6 +183,34 @@ ${iterator_types.incomplete_decl(T.entity.iterator)}
...
+ */
+
+${c_doc('langkit.event_handler_type')}
+typedef void *${event_handler_type};
+
+${c_doc('langkit.event_handler_unit_requested_type')}
+typedef void (*${event_handler_unit_requested_type})(
+   void *data,
+   ${analysis_context_type} context,
+   ${text_type} *name,
```
pmderodat, line 197: Unit names are filenames, and we represent them with bytes in the Ada and C APIs, so this should be `const char *` (like in `get_analysis_unit_from_file`)

Alternatively, IIRC we do create analysis units even when the file is not found, so we could pass directly that new analysis unit (which gives access to the filename) instead of just its file name. What do you think?

langkit/templates/c_api/header_c.mako
```diff
@@ -481,6 +510,23 @@ ${capi.get_name("get_versions")}(char **version, char **build_date);
     % endfor
 % endfor
 
+/*
+ * event handlers
```
pmderodat, line 514: Event*

langkit/templates/c_api/pkg_main_body_ada.mako
```diff
@@ -32,6 +32,39 @@ package body ${ada_lib_name}.Implementation.C is
    --  Avoid hiding from $.Lexer
    subtype Token_Data_Type is Common.Token_Data_Type;
 
+   ---------------------
+   -- Signal handlers --
```
pmderodat, line 36: Event*

langkit/templates/c_api/pkg_main_body_ada.mako
```diff
@@ -135,11 +169,12 @@ package body ${ada_lib_name}.Implementation.C is
...
-             File_Reader   => Unwrap_Private_File_Reader (File_Reader),
-             Unit_Provider => Unwrap_Private_Provider (Unit_Provider),
-             With_Trivia   => With_Trivia /= 0,
-             Tab_Stop      => Natural (Tab_Stop));
+            (Charset        => C,
+             File_Reader    => Unwrap_Private_File_Reader (File_Reader),
+             Unit_Provider  => Unwrap_Private_Provider (Unit_Provider),
+             Event_Handler => Unwrap_Private_Event_Handler (Event_Handler),
+             With_Trivia    => With_Trivia /= 0,
+             Tab_Stop       => Natural (Tab_Stop));
```
pmderodat, line 177: Arrow alignment issue

langkit/templates/c_api/pkg_main_body_ada.mako
```diff
@@ -1068,6 +1103,47 @@ package body ${ada_lib_name}.Implementation.C is
...
+      Clear_Last_Exception;
+      declare
+         Result : constant Internal_Event_Handler_Access :=
+           new C_Event_Handler'
+             (Ada.Finalization.Limited_Controlled with
+              Ref_Count               => 1,
+              Data                    => Data,
+              Destroy_Func            => Destroy_Func,
+              Unit_Requested_Func     => Unit_Requested_Func,
+              Unit_Parsed_Func        => Unit_Parsed_Func);
```
pmderodat, line 1122: Too much indentation for arrows?

langkit/templates/c_api/pkg_main_spec_ada.mako
```diff
@@ -529,6 +560,33 @@ private package ${ada_lib_name}.Implementation.C is
       ctx.ext('analysis', 'c_api', 'file_readers', 'spec')
    )}
 
+   ---------------------
+   -- Signal handlers --
```
pmderodat, line 564: Event* (I noticed other occurences)

langkit/templates/pkg_analysis_body_ada.mako
```diff
@@ -63,6 +63,16 @@ package body ${ada_lib_name}.Analysis is
    -- Do_Release --
    ----------------
 
+   procedure Do_Release
+     (Self : in out Event_Handler_Interface'Class) is
```
pmderodat, line 67: Could fit in one single line

langkit/templates/pkg_analysis_body_ada.mako
```diff
@@ -99,13 +123,16 @@ package body ${ada_lib_name}.Analysis is
          Wrap_Public_File_Reader (File_Reader);
       UP     : Internal_Unit_Provider_Access :=
          Wrap_Public_Provider (Unit_Provider);
+      SH     : Internal_Event_Handler_Access :=
```
pmderodat, line 126: EH*

langkit/templates/pkg_analysis_spec_ada.mako
```diff
@@ -109,6 +109,45 @@ package ${ada_lib_name}.Analysis is
...
+   type Event_Handler_Interface is interface;
+   ${ada_doc('langkit.event_handler_type')}
+
+   procedure Unit_Requested_Callback
+     (Self               : Event_Handler_Interface;
+      Context            : Analysis_Context'Class;
+      Name               : Text_Type;
+      From               : Analysis_Unit'Class;
+      Found              : Boolean;
+      Is_Not_Found_Error : Boolean) is null;
```
pmderodat, line 125: Could you add documentation for this callback and the other one?

langkit/templates/pkg_analysis_spec_ada.mako
```diff
@@ -109,6 +109,45 @@ package ${ada_lib_name}.Analysis is
...
+
+   procedure Unit_Parsed_Callback
+     (Self     : Event_Handler_Interface;
+      Context  : Analysis_Context'Class;
+      Unit     : Analysis_Unit'Class;
+      Reparsed : Boolean) is null;
+
+   procedure Release (Self : in out Event_Handler_Interface) is abstract;
+   --  Actions to perform when releasing resources associated to Provider
```
pmderodat, line 134: Self*

langkit/templates/pkg_analysis_spec_ada.mako
```diff
@@ -109,6 +109,45 @@ package ${ada_lib_name}.Analysis is
...
+     (Event_Handler_Interface'Class, Do_Release);
+
+   subtype Event_Handler_Reference is Event_Handler_References.Ref;
+   No_Event_Handler_Ref : Event_Handler_Reference renames
+      Event_Handler_References.Null_Ref;
+
+   function Create_Event_Handler_Reference
+     (Handler : Event_Handler_Interface'Class) return Event_Handler_Reference;
+   --  Simple wrapper around the GNATCOLL.Refcount API to create unit provider
+   --  references.
```
pmderodat, line 149: Please update the doc

langkit/templates/pkg_implementation_body_ada.mako
```diff
@@ -533,6 +554,11 @@ package body ${ada_lib_name}.Implementation is
          Unit.Charset := Actual_Charset;
       end if;
 
+      if Context.Event_Handler /= null then
+         Context.Event_Handler.Unit_Parsed_Callback
+           (Context, Unit, Reparse);
```
pmderodat, line 559: A single line could hold this

langkit/templates/pkg_implementation_spec_ada.mako
```diff
@@ -1336,6 +1336,41 @@ private package ${ada_lib_name}.Implementation is
...
+   ---------------------------------------
+   -- Signal handler internal interface --
+   ---------------------------------------
+
+   type Internal_Event_Handler is limited interface;
+   type Internal_Event_Handler_Access is
+      access all Internal_Event_Handler'Class;
+
+   procedure Inc_Ref (Self : in out Internal_Event_Handler) is abstract;
+   ${ada_doc('langkit.unit_provider_inc_ref', 3)}
```
pmderodat, line 1348: Wrong doc (same below)

testsuite/tests/misc/event_handler/extensions/src/libfoolang-implementation-extensions.adb
```diff
@@ -0,0 +1,25 @@
+with Langkit_Support.Text; use Langkit_Support.Text;
+
+package body Libfoolang.Implementation.Extensions is
+
+   function Foo_Node_P_Get_Unit
+     (Node           : Bare_Foo_Node;
+      Name           : Symbol_Type) return Internal_Unit
```
pmderodat, line 7: Too much alignment (same in spec)

testsuite/tests/misc/event_handler/extensions/python
```diff
@@ -0,0 +1,14 @@
+## vim: ft=makopython
```
pmderodat, line 1: Unused file (event handlers are not bound to the Python API anyway)

testsuite/tests/misc/event_handler/main.py
```diff
@@ -0,0 +1,73 @@
+import sys
```
pmderodat, line 1: Unused file (event handlers are not bound to the Python API anyway)

testsuite/tests/misc/event_handler/support.ads
```diff
@@ -0,0 +1,27 @@
+with Langkit_Support.Text;         use Langkit_Support.Text;
+with Libfoolang.Analysis; use Libfoolang.Analysis;
+
+package Support is
+   type Sig_Handler is new Libfoolang.Analysis.Event_Handler_Interface
```
pmderodat, line 5: Event*

testsuite/tests/misc/event_handler/test.py
```diff
@@ -0,0 +1,35 @@
...
+"""
+
+from langkit.dsl import ASTNode, AnalysisUnit, Symbol
+from langkit.expressions import langkit_property
+
+from utils import build_and_run
+
+
+class FooNode(ASTNode):
+    pass
```
pmderodat, line 12: Extra `pass`

testsuite/tests/misc/event_handler/test.py
```diff
@@ -0,0 +1,35 @@
...
+        pass
+
+
+class Example(FooNode):
+    token_node = True
+
+
+build_and_run(
+    lkt_file="expected_concrete_syntax.lkt",
+    generate_unparser=True,
```
pmderodat, line 31: Is this intentional? (no apparent need for unparsers in this test)

testsuite/tests/misc/event_handler/direct-decoding-error.txt
```diff
@@ -0,0 +1 @@
+éxample
```
pmderodat, line 1: Unused file

testsuite/tests/misc/event_handler/direct-bad-charset.txt
```diff
@@ -0,0 +1 @@
+example
```
pmderodat, line 1: Unused file

langkit/templates/c_api/pkg_main_spec_ada.mako
```diff
@@ -186,6 +186,36 @@ private package ${ada_lib_name}.Implementation.C is
       with Convention => C;
    ${ada_c_doc('langkit.file_reader_read_type', 3)}
 
+   ---------------------
+   -- Signal handlers --
+   ---------------------
+
+   type ${event_handler_type} is new System.Address;
+   ${ada_c_doc('langkit.file_reader_type', 3)}
```
pmderodat, line 194: Wrong doc

langkit/compiled_types.py
```diff
@@ -84,11 +84,21 @@ def template_extensions(ctx):
         'symbol_type':           T.Symbol.c_type(capi).name,
```
pmderodat, line 84: (for the commit message): Could you add a high level description of this new mechanism? We also need some user-facing doc for this.

pmderodat (changes requested):
Thanks for adding this feature! I have mostly minor requests for improvements and some doc is missing, but otherwise we’re good!

langkit/documentation.py
raph-amiard: aaah right, I wrote the docs and did the renaming signal -> event afterwards, so not surprising...

langkit/templates/c_api/header_c.mako
raph-amiard: > Alternatively, IIRC we do create analysis units even when the file is not found, so we could pass directly that new analysis unit (which gives access to the filename) instead of just its file name. What do you think?

The "unit requested" signal is supposed to happen *before* the unit is actually parsed, so I don't think it would make sense passing a unit here.

raph-amiard: Actually this isn't necessarily a file name here: Since for the moment the langkit user has the job of adding the callbacks, and will decide what to pass here, he could decide to pass the abstract language level name (`Ada.Directories`) or the filename (`a-direct.ads`).

The langkit test uses the first way, and libadalang the second, and depending on the language, I think both could make sense.

> Alternatively, IIRC we do create analysis units even when the file is not found, so we could pass directly that new analysis unit (which gives access to the filename) instead of just its file name. What do you think?

We would lose the ability to pass the abstract name, but we could pass both, and enforce that the `name` parameter must be the abstract name.

But to be honest, while I'm not sure this interface is perfect yet, I need to do a bit of thinking on the LAL side, and maybe a bit of cleaning up in env hooks & all, to see how this all could be simplified, and I'd rather revisit this after if that's OK with you.

pmderodat: I see, thanks for the rationale. Could you then add a warning in the doc of this callback? (very briefly describing the probable evolution would also be nice) In any case, enhancing the current doc to give semantics (or the lack of semantics) for this `name` argument looks necessary.

langkit/templates/pkg_analysis_spec_ada.mako
```diff
@@ -109,6 +109,63 @@ package ${ada_lib_name}.Analysis is
...
+   --  ``Context``.
+   --
+   --  ``Name`` is the name of the requested unit.
+   --
+   --  ``From`` is the unit from which the unit was requested.
+   --
+   --  ``Found`` indicates whether the requested unit was found or not.
+   --
+   --  ``Is_Not_Found_Error`` indicates whether the fact that the unit was not
+   --  found is an error or not.
```
pmderodat, line 136: Sorry, I should have made this explicit in my previous comment: what about using the `ada_doc` helper to avoid duplication with docs from `langkit.documentation`?

support/langkit_support-adalog-abstract_relation.adb
```diff
@@ -23,6 +23,7 @@
 
 with Ada.Text_IO; use Ada.Text_IO;
 with Ada.Unchecked_Deallocation;
+with Ada.Directories;
```
pmderodat, line 26: Wrong commit?

testsuite/tests/misc/event_handler/main.adb
```diff
@@ -0,0 +1,35 @@
...
+
+begin
+   Put_Line ("main.adb: Starting...");
+
+   --  Create a context with our file reader
+
+   declare
+      My_Sig_Handler : constant Event_Handler_Reference :=
+         Create_Event_Handler_Reference
+           (Sig_Handler'(null record));
```
pmderodat, line 22: This probably won’t build (`Event_Handler`)

There are other occurences (also missing message updates)

pmderodat (changes requested):

langkit/templates/pkg_analysis_spec_ada.mako
raph-amiard: It's not possible because only callbacks types are documented in langkit.documentation, not callback functions. You did the exact same thing for the file reader btw.

pmderodat: I did that for file readers because the file reader interface is defined in Langkit_Support. Here, everything is in Mako templates, so this should be possible, no?

langkit/templates/c_api/header_c.mako
raph-amiard: The problem with this is that this doc is going to be propagated in the generated bindings, where the semantic is fixed. The most important use case being Libadalang, where it is indeed a file name at the moment.

langkit/documentation.py
```diff
@@ -686,6 +686,69 @@ def instantiate_templates(doc_dict: Dict[str, str]) -> DocDatabase:
         buffer for a requested filename.
     """,
 
+    #
+g   # Event handlers
```
pmderodat, line 690: Typo

langkit/templates/c_api/header_c.mako
pmderodat: Ah, that’s annoying… we need to fix this eventually: can you leave a TODO?

pmderodat (changes requested):

langkit/templates/pkg_analysis_spec_ada.mako
raph-amiard: Except that the first part of the comment is not exactly the same: one refers to a callback, the other to a callback type

pmderodat: Ah yes, you are right! My bad, all is good, thanks.

(we should probably eventually find a way to share the common part, but that’s out of the scope of this change)

langkit/documentation.py
```diff
@@ -686,6 +686,69 @@ def instantiate_templates(doc_dict: Dict[str, str]) -> DocDatabase:
...
+        to ``${capi.get_name('dec_ref_event_handler')}``.
+
+        Pass as ``data`` a pointer to hold your private data: it will be passed
+        to all callbacks below.
+
+        ``destroy`` is a callback that is called by
+        ``${capi.get_name('dec_ref_event_handler')}`` to leave a chance to
+        free resources that ``data`` may hold.
+
+        ``unit_requested`` is a callback that will be called when an unit is
```
pmderodat, line 703: a unit* (same 2 times below)

pmderodat (approved):

PR #525, opened 2021-08-12
Public API enhancements

raph-amiard:


langkit/templates/pkg_analysis_body_ada.mako
```diff
@@ -937,6 +937,38 @@ package body ${ada_lib_name}.Analysis is
...
+   -----------------
+
+   function First_Child
+     (Node : ${root_entity.api_name}'Class) return ${root_entity.api_name} is
+   begin
+      Check_Safety_Net (Node);
+
+      if Node.Children_Count = 0 then
+         return No_${root_entity.api_name};
+      else
```
Roldak, line 951: This is not necessary, as the call to `Child` below already return none in that case (at least that's what the doc mentions). Same below!

Roldak (changes requested):
See my comment. Otherwise LGTM

Roldak (approved):

PR #527, opened 2021-09-08
Update langkit tutorial

thvnx:
This patchs updates the langkit tutorial, starting from the `Getting
Started` section. It consists of updating the new file/path names used
by langkit when creating a project and switching some Python code from
version 2 to 3.

It also adds a new file, `kalint.py` to help the reader to correctly
assemble all the pieces of the langit toy interpreter.

### DIFF
--- a/doc/tutorial.rst
+++ b/doc/tutorial.rst
@@ -48,7 +48,7 @@ directory and run:
 
 .. code-block:: text
 
-    $ create-project.py Kaleidoscope
+    $ scripts/create-project.py Kaleidoscope
 
 This will create a ``kaleidoscope`` directory, a dummy language specification
 (lexer and parser) as well as a ``manage.py`` script that will help you to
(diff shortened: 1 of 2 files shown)


doc/tutorial.rst
```diff
@@ -757,14 +766,14 @@ Our top-level code looks like this:
     def print_error(filename, sloc_range, message):
         line = sloc_range.start.line
         column = sloc_range.start.column
-        print >> sys.stderr, 'In {}, line {}:'.format(filename, line)
+        print ('In {}, line {}:'.format(filename, line), sys.stderr)
```
Roldak, line 769: shouldn't it be `print("...", file=sys.stderr)` instead of `print("...", sys.stderr)` to avoid the `<_io.TextIOWrapper name='<stderr>' mode='w' encoding='utf-8'>` stuff, as we saw together? (same below)

doc/tutorial.rst
```diff
@@ -644,18 +651,20 @@ code?
...
+``build/lib/relocatable/dev/libkaleidoscopelang`` and the
 ``build/lib/libkaleidoscopelang.relocatable`` directories is in your
-``LD_LIBRARY_PATH`` (on most Unix, ``DYLD_FALLBACK_LIBRARY_PATH`` on Darwin,
-adapt for Windows) and that the ``build/python/libkaleidoscopelang.py`` is
-reachable from Python (add ``build/python`` in your ``PYTHONPATH`` environment
-variable).
+``LD_LIBRARY_PATH`` (on most Unix, ``DYLD_FALLBACK_LIBRARY_PATH`` on
+Darwin, adapt for Windows) and that the
+``build/python/libkaleidoscopelang/__init__.py`` is reachable from
+Python (add ``build/python/libkaleidoscopelang`` in your
```
Roldak, line 666: I'm surprised that one should add `build/python/libkaleidoscopelang` to PYTHONPATH instead of `build/python`, could you double check that please?

Roldak (changes requested):

doc/tutorial.rst
thvnx: correct, I forgot to fix these ones!

doc/tutorial.rst
thvnx: double checked:
```
$ ls build/python/
libkaleidoscopelang  setup.py
```

thvnx: my bad, you're right!

Roldak (approved):

PR #531, opened 2021-09-13
U913-020: add parent and node_env Node's builtins to LKT prelude

thvnx:
Adding parent and node_env methods to Node specification in order to
make the following test pass:
testsuite/tests/structs/foreign_env_md/expected_concrete_syntax.lkt.

To do so, a new class has also been introduced: LexicalEnv, as well as
it's get method.

### DIFF
--- a/contrib/lkt/extensions/src/liblktlang-implementation-extensions.adb
+++ b/contrib/lkt/extensions/src/liblktlang-implementation-extensions.adb
@@ -36,7 +36,12 @@ package body Liblktlang.Implementation.Extensions is
      "@builtin struct Char {}" & ASCII.LF &
      "@builtin struct String " &
      "implements Sized, Indexable[Char], Iterator[Char] {}" & ASCII.LF &
+     "@builtin class LexicalEnv {" & ASCII.LF &
+     "    @builtin fun get(symbol : Symbol): Array[Node]" & ASCII.LF &
+     "}" & ASCII.LF &
      "@builtin class Node {" & ASCII.LF &
+     "    @builtin @property fun parent(): Node" & ASCII.LF &
+     "    @builtin fun node_env(): LexicalEnv" & ASCII.LF &
      "}" & ASCII.LF &
      "@builtin trait TokenNode {" & ASCII.LF &
      "    @builtin @property fun symbol(): Symbol" & ASCII.LF &
--- a/testsuite/tests/contrib/lkt_semantic/builtin_methods/test.lkt
+++ b/testsuite/tests/contrib/lkt_semantic/builtin_methods/test.lkt
@@ -1,0 +1,11 @@
+class Name : Node implements TokenNode {
+
+    fun parent_node (): Node = node.parent
+
+    fun parent_env (): LexicalEnv = node.parent.node_env()
+    fun parent_env (): LexicalEnv = node.parent_node().node_env()
+
+    fun parent_env_node (): Node = node.parent.node_env().get(node.symbol)?(0)
+    fun parent_env_node (): Node = node.parent_env().get(node.symbol)?(0)
+
+}
--- a/testsuite/tests/contrib/lkt_semantic/builtin_methods/test.out
+++ b/testsuite/tests/contrib/lkt_semantic/builtin_methods/test.out
@@ -1,0 +1,206 @@
+Resolving test.lkt
+==================
+Id   <RefId "Node" test.lkt:1:14-1:18>
+     references <ClassDecl "Node" __prelude:27:10-30:2>
+
+Id   <RefId "TokenNode" test.lkt:1:30-1:39>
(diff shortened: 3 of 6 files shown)


contrib/lkt/extensions/src/liblktlang-implementation-extensions.adb
```diff
@@ -36,7 +36,12 @@ package body Liblktlang.Implementation.Extensions is
      "@builtin struct Char {}" & ASCII.LF &
      "@builtin struct String " &
      "implements Sized, Indexable[Char], Iterator[Char] {}" & ASCII.LF &
+     "@builtin class LexicalEnv {" & ASCII.LF &
```
raph-amiard, line 39: You have a regression in the testsuite due to the fact that entities have moved. Run the testsuite with the `-r` option to change the test baseline to the test's actual output.

`./manage.py test <test_name> -r`

raph-amiard (changes requested):

contrib/lkt/extensions/src/liblktlang-implementation-extensions.adb
thvnx: Thanks for the hint!

testsuite/tests/contrib/lkt_syntax/builtin_methods.lkt
```diff
@@ -0,0 +1,11 @@
+class Name : Node implements TokenNode {
```
pmderodat, line 1: IIUC, this new file is unused during the execution of the `contrib/lkt_syntax` testcase… what about adding a new test in `contrib/lkt_semantic` instead?

Also, the commit message mentions fixing the `structs/foreign_env_md` test, but this testcase already passes, so it’s not clear what is fixed. Maybe you intended to pass `lkt_semantic_checks=True` to its `build_and_run`?

pmderodat (changes requested):
Thanks and congratulations for your first take in Lkt, Laurent! I only have remarks regarding testing…

testsuite/tests/contrib/lkt_syntax/builtin_methods.lkt
thvnx: Is it really unused? I'm still discovering how the test suite works but I thought that just adding a new file is enough. @raph-amiard, what do you think about the `lkt_syntax` vs `lkt_semantic` folder for this test.

Sorry, the commit message wasn't clear enough. You're right, this test already pass but it doesn't when you use the new in-development version of the langkit language. This task was given to me as an exercice to make me discover how the new LKT frontend is supposed to work, and to reduce the gap between the current working version of langkit and the next one. In that scope, I had to make this test pass with the future version of LKT.

Hope it helps!

pmderodat: > Is it really unused? I'm still discovering how the test suite works but I thought that just adding a new file is enough.

No problem. :-) For most (all?) tests, the entry point is the `test.py` file, so https://github.com/AdaCore/langkit/blob/master/testsuite/tests/contrib/lkt_syntax/test.py in this case. The real work happens in the `build_and_run` function (defined here: https://github.com/AdaCore/langkit/blob/master/testsuite/python_support/utils.py#L184), and we can see there that it just compiles the `expected_concrete_syntax.lkt` source file, which leads me to think that `builtin_methods.lkt` is ignored.

> Sorry, the commit message wasn't clear enough. You're right, this test already pass but it doesn't when you use the new in-development version of the langkit language. This task was given to me as an exercice to make me discover how the new LKT frontend is supposed to work, and to reduce the gap between the current working version of langkit and the next one. In that scope, I had to make this test pass with the future version of LKT.

Yep, we’re on the same page about this. :-) Now, even though the transition to Lkt (as opposed to the Python DSL) is still in the works, we try in the testsuite to excercize everything that is implemented. This is why `build_and_run` has the two `types_from_lkt` and `lkt_semantic_checks` arguments: by default we just make sure that the Lkt unparsing (go check with Raphaël and Romain if you don’t know what it is yet) yields what is expected (in `expected_concrete_syntax.lkt`), but these two arguments make it possible to respectively use definitions in the Lkt source file to generate the analysis library or just run semantic checks on that source.

Here, to check that the new builtin declarations you are adding have the intended effect, I’m suggesting to edit the `structs/foreign_env_md` testcase so that semantic checks are enabled, in the hope that these semantic check work on the properties using `.parent` and lexical envs, and thus work with your changes (i.e. they would fail without). We try to do this for every feature we add to Lkt.

thvnx: Thanks for the explanation, I guess it's all good now?

raph-amiard (approved):
Looking good, just wait for the CI to finish running and confirm everything is good and you can merge.

thvnx:
Thanks again!

PR #532, opened 2021-09-14
RA22-015: Fix call decl/type resolution

raph-amiard:
This fixes several bugs in the call decl/type resolution mechanism

* Generally make `called_decl` only return a called decl when there is
  one for the specific callexpr. Before, for nested call expressions
  like `a()()()` it would return the sub call-expression called decl.

* Make `expr_context_free_type` compute the type even when there is no
  called declaration

### DIFF
--- a/contrib/lkt/language/parser.py
+++ b/contrib/lkt/language/parser.py
@@ -2098,11 +2098,17 @@ class SimpleTypeRef(TypeRef):
 
     @langkit_property(return_type=T.SemanticResult.array)
     def check_correctness_pre():
+
         d = Var(Entity.type_name.referenced_decl)
+
         return d.result_ref.then(
-            lambda d: d.cast(T.TypeDecl).then(
-                lambda _: No(T.SemanticResult.array),
-                default_val=Entity.error(S("Invalid type reference")).singleton
+            lambda d: d.match(
+                # The type ref references a type decl: return no error
+                lambda _=T.TypeDecl: No(T.SemanticResult.array),
+
+                # Not a type decl: return an error that the type reference is
+                # invalid.
+                lambda _: [Entity.error(S("Invalid type reference"))]
             )
         )
 
--- a/contrib/lkt/language/parser.py
+++ b/contrib/lkt/language/parser.py
@@ -1933,10 +1933,21 @@ class FunDecl(UserValDecl):
     call_scope = Property(Entity.children_env)
 
     @langkit_property()
+    def is_property():
+        """
+        Returns whether this function decl defines a property or not.
+
+        A property is a function that has the @property annotation and has no
+        parameters. As a result of this annotation, it will be callable without
+        parens.
+        """
+        return Entity.full_decl.has_annotation('property')
+
+    @langkit_property()
     def get_type(no_inference=(T.Bool, False)):
         ignore(no_inference)
         return If(
-            Entity.full_decl.has_annotation('property'),
+            Entity.is_property,
 
             Entity.return_type.designated_type,
 
@@ -2509,20 +2520,50 @@ class CallExpr(Expr):
     @langkit_property(public=True)
     def called_decl():
         """
-        Return the declaration that is called by this call expression.
+        Return the declaration that is called by this call expression, if there
+        is one that can be statically determined.
         """
-        refd_decl = Var(Entity.name.referenced_decl.then(
-            lambda sr: sr.result_ref
+
+        # We use this to resolve formals rather than just the type of the
+        # ``name`` of the callexpr, for two language features:
+        #
+        # * Generic functions: We need to statically resolve them in order to
+        #   allow inference in the implicit instantiation.
+        #
+        # * Named parameters: Since function types don't contain parameter
+        #   names, if we want to allow named parameters we need to statically
+        #   resolve the target of the call when possible.
+
+        refd_decl = Var(If(
+
+            # If the name of this callexpr is a callexpr itself, there cannot
+            # be a statically known referenced decl, since whatever is
+            # referenced is the result.
+            Not(Entity.name.is_a(CallExpr)),
+
+            Entity.name.referenced_decl.then(
+                lambda sr: sr.result_ref
+            ).then(lambda decl: If(
+                decl.cast(T.FunDecl)._.is_property, No(T.Decl.entity), decl
+            )),
+
+            No(T.Decl.entity)
         ))
 
-        # Implement special resolution for calls to objects via __call__
+        # Implement special resolution for calls to objects via __call__.
+        # NOTE: We do this resolution here because in case where the object is
+        # statically resolvable, the __call__ method might have keyword
+        # arguments, see comment above.
         called_decl = Var(
             refd_decl.then(lambda refd_decl: refd_decl.match(
+
                 # Don't try to look for __call__ on function decls
                 lambda _=FunDecl: refd_decl,
+
                 lambda v=BaseValDecl:
                 v._.get_type._.get_fun('__call__')
                 .then(lambda a: a, default_val=refd_decl),
+
                 lambda _: refd_decl,
             ))
         )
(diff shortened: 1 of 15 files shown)


thvnx:
Note that you have a regression (`testsuite/tests/contrib/lkt_semantic/array_list_methods/test.lkt`):

```
# Here, self doesn't exist. Since we're going to try to resolve the type of
# `12` first, it's going to need to resolve `self, which doesn't exist. This
# shouldn't crash.
@invalid fun test_astlist_indexing(a: ASTList[String]): String = self(12)
```

Looks like `self(12)` type was inferred from `<StructDecl "String" __prelude:23:10-23:76>`, but not anymore. Not sure to understand if it's expected though.

Is that correct?

raph-amiard:
> Looks like self(12) type was inferred from <StructDecl "String" __prelude:23:10-23:76>, but not anymore. Not sure to understand if it's expected though.

Yep that's what was happening. More precisely, bottom up type inference was infering the type from the function signature (Since it returns a string). 

It's a side effect of my change that doesn't matter: we don't really care about inferring this type in incorrect code, since technically this expression (`self(12)`) doesn't really have a type. I'll update the baselines.

thvnx (approved):
LGTM

Roldak (approved):

contrib/lkt/language/parser.py
```diff
@@ -1932,11 +1932,22 @@ class FunDecl(UserValDecl):
 
     call_scope = Property(Entity.children_env)
 
+    @langkit_property()
+    def is_property():
+        """
+        Returns whether this function decl defines a property or not.
```
pmderodat, line 1938: This should be return*

contrib/lkt/language/prelude.lkt
```diff
@@ -35,7 +35,7 @@ trait Iterator {
 
 @builtin generic[T]
 struct Array implements Sized, Indexable[T], Iterator[T] {
-    @builtin @property fun to_iterator(): Iterator[T]
+    @builtin fun to_iterator(): Iterator[T]
```
pmderodat, line 38: The commit message is not very informative: why is it a fix? Looking a the change, it looks we just want to change the signature of `to_iterator` (API change, not a bug fix) Also I don’t understand why this is needed even looking at the next commits: please add a testcase and/or expand the commit message.

contrib/lkt/language/parser.py
```diff
@@ -2509,20 +2520,50 @@ def called_object_type():
...
-        Return the declaration that is called by this call expression.
+        Return the declaration that is called by this call expression, if there
+        is one that can be statically determined.
         """
-        refd_decl = Var(Entity.name.referenced_decl.then(
-            lambda sr: sr.result_ref
+
+        # We use this to resolve formals rather than just the type of the
+        # ``name`` of the callexpr, for two language features:
+
```
pmderodat, line 2535: For consistency, can you add empty comments between paragraphs?

contrib/lkt/language/parser.py
```diff
@@ -2509,20 +2520,50 @@ def called_object_type():
...
+        # * Generic functions: We need to statically resolve them in order to
+        # allow inference in the implicit instantiation.
```
pmderodat, line 2537: Can you fix indentation (“allow” aligned with “Generic”, same below)

contrib/lkt/language/parser.py
```diff
@@ -2509,20 +2520,50 @@ def called_object_type():
...
+
+        # * Named parameters: Since function types don't contain parameter
+        # names, if we want to allow named parameters we need to statically
+        # resolve the target of the call when possible.
```
pmderodat, line 2541: I have the feeling we should allow named arguments even when the target of the call is not resolvable (for instance when dealing with property references, which don’t exist right now, but could be useful in the future). WDYT? (Ok to do that later of course)

testsuite/tests/contrib/lkt_semantic/advanced_calls/test.lkt
```diff
@@ -0,0 +1,19 @@
+class MyCallable {
+    fun __call__(a : Int): Bool
```
pmderodat, line 2: Can you adjust the formatting according to the style document?

testsuite/tests/structs/foreign_env_md/test.yaml
```diff
@@ -1 +1,3 @@
 driver: python
+control:
+    - [XFAIL, "True", 'Pending resolution of U920-002']
```
pmderodat, line 3: This ticket has been closed: update needed?

contrib/lkt/language/parser.py
```diff
@@ -2579,6 +2579,23 @@ def called_decl():
             ),
             # If not a generic decl, fall back on the initially found decl
             default_val=called_decl
+        )._.match(
+            # Here, if we found a declaration but it's not one of the
+            # statically callable declarations (so FunDecl and TypeDecl), we
```
pmderodat, line 2590: I have trouble understanding this, I feel this is ambiguous: are FunDecl and TypeDecl statically callable declarations? If so I would suggest to write “i.e.” instead of “so”, and if not, write “so it’s one of …”

contrib/lkt/language/parser.py
```diff
@@ -2579,6 +2579,23 @@ def called_decl():
...
+            # determine a declaration, like in the example below::
+            #
+            #    val a : (Int) -> Int = null
+            #    val b : Int = a(12)
+            #
+            # That's because in language semantic terms, the actual called
+            # function is determined dynamically in the above case.
+            #
+            # TODO: Introduce a base class, so we don't have to do this ugly
+            # match?
```
pmderodat, line 2601: Can you clarify? (a base class for what? how woud it help to avoid the match? I don’t understand why it’s ugly, …)

testsuite/tests/contrib/lkt_semantic/call_non_static_decl/test.lkt
```diff
@@ -0,0 +1,4 @@
+# Call a non static declaration
```
pmderodat, line 1: Please update for the coding style

testsuite/tests/contrib/lkt_semantic/call_non_callable/test.lkt
```diff
@@ -0,0 +1,7 @@
+# Error for trying to call a non callable
+
+val a : (Int) -> Int = null
```
pmderodat, line 3: Please update the coding style

pmderodat (changes requested):

contrib/lkt/language/parser.py
raph-amiard: As discussed live: yes, let's do that (long term)

contrib/lkt/language/parser.py
raph-amiard: It's ugly because if we had a base class we could just do `.cast(StaticCallableDecl)` or sth

contrib/lkt/language/prelude.lkt
raph-amiard: Commit msg has been updated as follows: 

>>> 
Will be needed for an upcoming fix on call resolution. For the moment
to_iterator is unparsed as a regular function call in dsl_unparse, but
in the prelude it's a property, so unify towards it being a function.
Some discrepancies (see entity_iterator test) are ignored for the moment
due to a hole in type resolution.

contrib/lkt/language/parser.py
```diff
@@ -2270,21 +2270,36 @@ def check_correctness_pre():
...
+                lambda generic_decl: Entity.infer_actuals(
+                    generic_decl,
+                    Entity.expected_type
+                ).then(lambda iir: iir.error.then(lambda e: e.singleton)),
+
+            )
+
+            # Handle errors for when we have a call to an object with no
+            # declaration, so we need to rely on the object_type, and the
+            # object type is non callable.
```
pmderodat, line 2303: I have trouble understanding this comment: what’s an object with no declaration? Do you mean when we have a call that we can’t resolve to a declaration?

contrib/lkt/language/prelude.lkt
pmderodat: Perfect, thanks!

contrib/lkt/language/parser.py
pmderodat: :cry: okay

contrib/lkt/language/parser.py
pmderodat: This one does not look resolved: unresolving it

contrib/lkt/language/parser.py
```diff
@@ -2589,6 +2589,23 @@ def called_decl():
...
             # If not a generic decl, fall back on the initially found decl
             default_val=called_decl
+        )._.match(
+            # Here, if we found a declaration but it's not one of the
+            # statically callable declarations (FunDecl or TypeDecl), we
+            # return nothing, even though we might be able to statically
+            # determine a declaration, like in the example below::
+            #
+            #    val a : (Int) -> Int = null
+            #    val b : Int = a(12)
```
pmderodat, line 2605: Can you update according to the Lkt coding style?

testsuite/tests/contrib/lkt_semantic/call_non_callable/test.lkt
pmderodat: Updates missing

pmderodat (changes requested):

contrib/lkt/language/parser.py
raph-amiard: I don't understand what I did, but it should be fixed now

contrib/lkt/language/parser.py
raph-amiard: yes, amending

contrib/lkt/language/parser.py
pmderodat: Ping

pmderodat (approved):

PR #533, opened 2021-09-16
U914-006: emit error message on dependency not found

thvnx:
Properly emit an error message when the file pointed out by an
`import` statement is not found.

### DIFF
--- a/contrib/lkt/language/parser.py
+++ b/contrib/lkt/language/parser.py
@@ -507,6 +507,14 @@ class Import(LKNode):
         """
         return Self.internal_fetch_referenced_unit(Self.name.text)
 
+    @langkit_property(return_type=T.SemanticResult.array)
+    def check_correctness_pre():
+        return If(
+            Self.referenced_unit.root.is_null,
+            Entity.error(S("cannot find ").concat(Self.name.text)).singleton,
+            No(T.SemanticResult.array)
+        )
+
     env_spec = EnvSpec(
         do(Self.referenced_unit)
     )
--- a/testsuite/tests/contrib/lkt_semantic/invalid_import/test.lkt
+++ b/testsuite/tests/contrib/lkt_semantic/invalid_import/test.lkt
@@ -1,0 +1,1 @@
+import invalid_import
--- a/testsuite/tests/contrib/lkt_semantic/invalid_import/test.out
+++ b/testsuite/tests/contrib/lkt_semantic/invalid_import/test.out
@@ -1,0 +1,6 @@
+Resolving test.lkt
+==================
+test.lkt:1:1: error: cannot find invalid_import
+0 | import invalid_import
+  | ^^^^^^^^^^^^^^^^^^^^^
+
--- a/testsuite/tests/contrib/lkt_semantic/invalid_import/test.yaml
+++ b/testsuite/tests/contrib/lkt_semantic/invalid_import/test.yaml
@@ -1,0 +1,1 @@
+driver: lkt


contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -90,6 +90,8 @@ begin
             Unit : constant Analysis.Analysis_Unit
               := Ctx.Get_From_File (To_String (File_Name));
          begin
+            Unit.Populate_Lexical_Env;
```
Roldak, line 93: I would add a comment explaining why we explicitly call `Populate_Lexical_Env` here, instead of letting it be done automatically during the call to the `P_Check_Semantic` property.

contrib/lkt/extensions/src/liblktlang-implementation-extensions.adb
```diff
@@ -162,6 +162,14 @@ package body Liblktlang.Implementation.Extensions is
          Charset => "ascii",
          Reparse => False);
    begin
+      if Node.Kind = lkt_Import and I.AST_Root = null then
```
Roldak, line 165: I'm not sure there's any benefit in restraining the diagnostic emission to the case where `Node` is of kind `lkt_Import`. Do you have anything in mind? My intuition is that if we ever need to call this property from another kind of node, we will probably want the same kind of error handling (or it will be easy to refine the condition then).

Roldak (changes requested):

contrib/lkt/extensions/src/liblktlang-implementation-extensions.adb
thvnx: um... I just wanted to ensure that this specific diagnostic is only added for import nodes but if I understand what you're saying, it is possible to have this error on another node than a lkt_import kind one?

contrib/lkt/extensions/mains/lkt_toolbox.adb
thvnx: Done

contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -90,6 +90,12 @@ begin
             Unit : constant Analysis.Analysis_Unit
               := Ctx.Get_From_File (To_String (File_Name));
          begin
+            --  We explicitly call Populate_Lexical_Env here in order to be
+            --  able to print the diagnostics about invalid import statements
+            --  (U914-006). Indeed, waiting the call to P_Check_Semantics below
```
pmderodat, line 95: Except for temporary TODOs, I would suggest not to leave ticket numbers in comments: the codebase is public whereas tickets aren’t, so permanent comments/doc/… should be self-sufficient.

pmderodat (changes requested):

contrib/lkt/extensions/mains/lkt_toolbox.adb
thvnx: I see your point. I'll fix that!

contrib/lkt/extensions/src/liblktlang-implementation-extensions.adb
thvnx: ping @Roldak

pmderodat: I’ve just realized this violates the design of diagnostics, which are for lexing/parsing errors (as described in the error message from `lkt_resolve.py`), while here we have an error happening during semantic analysis.

@raph-amiard recently introduce a mechanism precisely to be notified when a file is missing (see event handlers, from U712-008). Unfortunately this mechanism is not yet available from the Python API.

Hm… what about re-implementing `lkt_resolve.py` in Ada, to use the proper abstraction for this?

raph-amiard: > I’ve just realized this violates the design of diagnostics, which are for lexing/parsing errors

I'm not sure this was ever a real design point, it's just that so far they've only been used for that. Now, you might have reasons to think it's better to keep them for that ?

> @raph-amiard recently introduce a mechanism precisely to be notified when a file is missing (see event handlers, from U712-008). Unfortunately this mechanism is not yet available from the Python API.

This mechanism unfortunately doesn't yet include the "origin" of the signal, ie. the import here, so wouldn't be enough even in Ada.

> Hm… what about re-implementing lkt_resolve.py in Ada, to use the proper abstraction for this?

We already have that, it's lkt_toolbox... TBH I thought all lkt tests were already using it ... Just realized it's using lkt_resolve.py. 

But it's not a great solution anyway: In Langkit we'll use lkt from python, so we need a solution that works in Python.

pmderodat: > > I’ve just realized this violates the design of diagnostics, which are for lexing/parsing errors
> 
> I'm not sure this was ever a real design point, it's just that so far they've only been used for that. Now, you might have reasons to think it's better to keep them for that ?

Yes: if you load unit A that tried to import unit B, the error message will go to unit A’s diagnostics. However if the loading of B is succesful, but this unit fails to load unit C, the diagnostic will go to unit B, so you won’t see looking at unit A that there is an error. So it looks like we need something global.

> This mechanism unfortunately doesn't yet include the "origin" of the signal, ie. the import here, so wouldn't be enough even in Ada.

I see, but that’s probably a sign that we should extend this API to include the origin.

> We already have that, it's lkt_toolbox... TBH I thought all lkt tests were already using it ... Just realized it's using lkt_resolve.py.
> 
> But it's not a great solution anyway: In Langkit we'll use lkt from python, so we need a solution that works in Python.

Ok… IMO this means we should tackle this and bind this API in Python. I’ll have a look at the feasability.

raph-amiard: Thinking about this further, I think this diagnostic should simply be emited by the `p_check_semantic`/`p_check_legality` properties of the DSL, as part of the normal error reporting process, and that it's pretty straightforward to do that ...

One thing that Romain raised is that on this error message we probably want to stop further analysis, or at  least display, since you're going to get a lot of garbage error messages, and Laurent's current solution has this benefit.

But this can probably be done if we really want to, I'll check to see how easy it would be

pmderodat: Ah yes, that’s even better! I forgot about the existing semantic checks machinery in Lkt… I’m much more in favor of that!

thvnx: Rework done!

pmderodat (approved):

Roldak (dismissed):

raph-amiard (approved):

PR #534, opened 2021-09-17
U916-032: Lkt: add more methods and types in builtins prelude

thvnx:


thvnx:
Depends on #533

thvnx:
> Depends on #533

Doesn't really depends on this one, should be passing, independently.

testsuite/tests/contrib/lkt_semantic/paren_expr/test.out
```diff
@@ -34,5 +34,5 @@ Expr <ParenExpr test.lkt:4:22-4:25>
      has type <StructDecl "Int" __prelude:1:10-1:23>
 
 Expr <BinOp test.lkt:4:16-4:25>
-     has type <EnumTypeDecl "Bool" __prelude:5:16-7:2>
```
raph-amiard, line 37: @Roldak @pmderodat I don't know what you think but I wonder if we shouldn't hide slocs here like we do in Ada nameres for runtime entities

pmderodat: Agreed! JIC: we should do this only in the `lkt_*` binary, not in the `Image` override in liblktlang, as having the sloc there is useful for debugging, i.e. we want to hide the sloc only for test outputs, for baseline stability.

raph-amiard: Ok cool. Let's do this for  further PRs. I'll approve this one

raph-amiard (approved):

thvnx:
Thanks!

PR #535, opened 2021-09-20
U920-002: avoid mismatched type error on root node

thvnx:
As the LexicalEnv.get method always return an array of root nodes while the root node is defined in user code, we need to turn the Node and LexicalEnv classes into generic traits and instantiate them using the actual root node.

To make this works, root node declaration has to be changed from `class FooNode : Node` to `class FooNode implements Node[FooNode]`, which required to refactor a bit almost all the lkt tests.

### DIFF
--- a/contrib/lkt/language/parser.py
+++ b/contrib/lkt/language/parser.py
@@ -137,15 +137,16 @@ class LKNode(ASTNode):
 
     @langkit_property(return_type=T.Decl.entity)
     def root_get(entity_name=T.Symbol):
-        return Self.unit.root.node_env.get_first(entity_name).cast(T.Decl)
+        return Self.unit.root.node_env.get_first(
+            entity_name).cast_or_raise(T.Decl)
 
     @langkit_property(return_type=T.TypeDecl.entity)
     def get_builtin_type(entity_name=T.Symbol):
-        return Self.root_get(entity_name).cast(T.TypeDecl)
+        return Self.root_get(entity_name).cast_or_raise(T.TypeDecl)
 
     @langkit_property(return_type=T.GenericDecl.entity)
     def get_builtin_gen_decl(entity_name=T.Symbol):
-        return Self.root_get(entity_name).cast(T.GenericDecl)
+        return Self.root_get(entity_name).cast_or_raise(T.GenericDecl)
 
     node_type = Property(
         Self.get_builtin_type('Node'), public=True,
--- a/contrib/lkt/language/parser.py
+++ b/contrib/lkt/language/parser.py
@@ -149,7 +149,7 @@ class LKNode(ASTNode):
         return Self.root_get(entity_name).cast_or_raise(T.GenericDecl)
 
     node_type = Property(
-        Self.get_builtin_type('Node'), public=True,
+        Self.get_builtin_gen_decl('Node'), public=True,
         doc="Unit method. Return the Node base class."
     )
 
--- a/langkit/dsl_unparse.py
+++ b/langkit/dsl_unparse.py
@@ -1286,6 +1286,7 @@ def emit_field(field):
 
 def type_name(type):
     from langkit.compiled_types import ASTNodeType, resolve_type
+    from langkit.compile_context import get_context
 
     type = resolve_type(type)
 
@@ -1315,6 +1316,11 @@ def type_name(type):
         return "{}.node".format(type.dsl_name)
     elif type.is_character_type:
         return 'Char'
+    elif type.is_lexical_env_type or type.is_analysis_unit_type:
+        return "{}[{}]".format(type.dsl_name,
+                               type_name(get_context().root_grammar_class))
+
+        return "".format
     else:
         return type.dsl_name
 
@@ -1350,7 +1356,7 @@ def emit_node_type(node_type):
         builtin_properties = node_type.builtin_properties()
 
         if node_type.is_root_node:
-            base_name = "Node"
+            traits.append(f'Node[{type_name(node_type)}]')
 
         abstract_qual = "@abstract " if node_type.abstract else ""
 
--- a/langkit/lkt_lowering.py
+++ b/langkit/lkt_lowering.py
@@ -53,7 +53,7 @@ def get_trait(decl: L.TypeDecl, trait_name: str) -> Optional[L.TypeDecl]:
     """
     for trait in decl.f_traits:
         trait_decl: L.TypeDecl = trait.p_designated_type
-        if trait_decl.f_syn_name.text == trait_name:
+        if trait_decl.p_name == trait_name:
             return trait_decl
     return None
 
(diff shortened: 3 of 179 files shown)


langkit/lkt_lowering.py
```diff
@@ -47,13 +47,17 @@
...
+              by_syn_name: bool = True) -> Optional[L.TypeDecl]:
     """
     Return the trait named ``trait_name`` on declaration ``decl``.
     """
     for trait in decl.f_traits:
         trait_decl: L.TypeDecl = trait.p_designated_type
-        if trait_decl.f_syn_name.text == trait_name:
+        if by_syn_name and trait_decl.f_syn_name.text == trait_name:
+            return trait_decl
+        elif trait_decl.p_name == trait_name:
```
pmderodat, line 60: I don’t understand why this hunk is necessary: the only case where `by_syn_name` is False is for the `Node` trait, and `f_syn_name.text` should always return "None" in this case. Could we remove this part of the patch?

testsuite/tests/contrib/lkt_semantic/builtin_methods/test.out
```diff
@@ -1,206 +1,169 @@
...
 
-Expr <DotExpr test.lkt:3:32-3:43>
-     has type <ClassDecl "Node" __prelude:27:10-30:2>
+test.lkt:3:32: error: Mismatched types: expected `Node`, got `Name`
+2 |     fun parent_node (): Node = node.parent
+  |                                ^^^^^^^^^^^
+
+test.lkt:5:24: error: Invalid type reference
+4 |     fun parent_env (): LexicalEnv = node.parent.node_env()
+  |                        ^^^^^^^^^^
```
pmderodat, line 30: Looks like the testcase needs more updates :-)

pmderodat (changes requested):
Thanks a lot for the progress in this area!

contrib/lkt/extensions/src/liblktlang-implementation-extensions.adb
```diff
@@ -36,12 +36,12 @@ package body Liblktlang.Implementation.Extensions is
...
      "implements Sized, Indexable[Char], Iterator[Char] {}" & ASCII.LF &
-     "@builtin class LexicalEnv {" & ASCII.LF &
-     "    @builtin fun get(symbol : Symbol): Array[Node]" & ASCII.LF &
+     "@builtin generic[NodeType] trait LexicalEnv {" & ASCII.LF &
+     "    @builtin fun get(symbol : Symbol): Array[NodeType]" & ASCII.LF &
      "}" & ASCII.LF &
-     "@builtin class Node {" & ASCII.LF &
-     "    @builtin @property fun parent(): Node" & ASCII.LF &
-     "    @builtin fun node_env(): LexicalEnv" & ASCII.LF &
+     "@builtin generic[ConcreteBaseNode] trait Node {" & ASCII.LF &
```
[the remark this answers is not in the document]
raph-amiard: `RootNode` SGTM!

contrib/lkt/language/parser.py
```diff
@@ -141,14 +141,14 @@ def root_get(entity_name=T.Symbol):
 
     @langkit_property(return_type=T.TypeDecl.entity)
     def get_builtin_type(entity_name=T.Symbol):
-        return Self.root_get(entity_name).cast(T.TypeDecl)
+        return Self.root_get(entity_name).cast_or_raise(T.TypeDecl)
```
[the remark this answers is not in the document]
raph-amiard: I agree with @pmderodat that this would go better in another commit

As for the testcase, not sure it's feasible nor useful: The reason we're using `cast_or_raise` here is because this is used in internal methods where we expect to return a type. Here, it made the `node_type` property return `None`, which is clearly not great, rather than raising an exception, which is better IMHO because it announces that underlying assumptions have been broken.

Since it's an internal property it's not easy to test.

langkit/lkt_lowering.py
raph-amiard: As discussed live, this (and the `by_syn_name` param) are unneeded AFAICT, you can just remove this branch and the param

contrib/lkt/extensions/src/liblktlang-implementation-extensions.adb
```diff
@@ -36,12 +36,12 @@ package body Liblktlang.Implementation.Extensions is
      "@builtin struct Char {}" & ASCII.LF &
```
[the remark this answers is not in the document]
thvnx: Much better! I edited the PR description and I'll amend the commit message accordingly.

testsuite/tests/contrib/lkt_semantic/builtin_methods/test.out
thvnx: Good catch!

raph-amiard: Yup, looks like you're going to have to modify `dsl_unparse.py` again :)

Here apparently you're going to need to replace `LexicalEnv` by `LexicalEnv[<name of root node type>]` everywhere.

I think that eventually we want to think of a shortcut way to refer to those things, like we discussed yesterday with @Roldak, but we need to think about this a little bit more, so in the meantime let's update the tests with the verbose syntax.

raph-amiard (changes requested):

contrib/lkt/extensions/src/liblktlang-implementation-extensions.adb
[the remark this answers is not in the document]
thvnx: Sure!

contrib/lkt/language/parser.py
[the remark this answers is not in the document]
thvnx: OK. I'll move them in another commit then! For the test, let me know if I should add one!

langkit/lkt_lowering.py
thvnx: OK, thanks for the explanation!

testsuite/tests/contrib/lkt_semantic/builtin_methods/test.lkt
```diff
@@ -1,4 +1,4 @@
-class Name : Node implements TokenNode {
+@abstract class Name implements Node[Name] {
```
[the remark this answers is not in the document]
thvnx: Should be fixed now.

testsuite/tests/contrib/lkt_semantic/builtin_methods/test.out
thvnx: Done, thanks for the help!

contrib/lkt/language/parser.py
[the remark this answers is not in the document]
pmderodat: Fair enough, I agree with Raphaël. Thanks! I suggest adding a justification to the commit message for such cases (here, what Raph said is good).

Shouldn’t we do `cast_or_raise`in `root_get` as well, by the way?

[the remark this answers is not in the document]
thvnx: Done. Also added `cast_or_raise` in `root_get`.

testsuite/tests/contrib/lkt_semantic/cast/test.lkt
```diff
@@ -1,4 +1,6 @@
-class RootNode : Node {}
+@abstract class CastNode {}
+
+class RootNode : CastNode {}
```
pmderodat, line 3: We don’t check this right now but the root node should implement Node all other classes should derive from it: could you fix accordingly?

langkit/dsl_unparse.py
```diff
@@ -1350,7 +1356,7 @@ def emit_node_type(node_type):
         builtin_properties = node_type.builtin_properties()
 
         if node_type.is_root_node:
-            base_name = "Node"
+            traits.append(f'Node[{type_name(node_type)}]')
```
pmderodat, line 1359: I can’t find an issue in the updated baselines, so I suppose this part is not exercized yet, but this looks wrong to me: shouldn’t we just use `type_name(node_type)` to designate the root node type? `Node[…]` looks just like an  instantiated trait (the one that the root node implements), which is different.

pmderodat (changes requested):

langkit/dsl_unparse.py
thvnx: Not sure to get your point here.

What we want now for the root node is to print it as a trait implementing `Node`, e.g.: `class FooNode implements Node[FooNode]` in place of `class FooNode : Node` for instance.

pmderodat: Ah yes, my bad! I thought the purpose of this part of the code was to emit a type reference to `node_type` instead of emitting its type declaration. So this is actually fine, sorry!

pmderodat (approved):
Congrats for this delicate change! LGTM

raph-amiard (approved):

PR #547, opened 2021-09-28
Lkt: PropertyError parsing support

thvnx:


testsuite/tests/properties/memoized_property_error/test.py
```diff
@@ -28,5 +28,9 @@ def prop2():
         return PropertyError(Bool, 'Explicit error 2')
 
 
-build_and_run(lkt_file='expected_concrete_syntax.lkt', ada_main='main.adb')
+build_and_run(lkt_file='expected_concrete_syntax.lkt', ada_main='main.adb',
+
+              # FIXME: internal errors without diagnostics when set to True,
```
pmderodat, line 33: Could you replace FIXME with TODO? (to be consistent with the rest of the codebase, and thus make it easier to spot such issues with a grep)

pmderodat (changes requested):
LGTM with a minor nit, thanks!

testsuite/tests/properties/memoized_property_error/test.py
thvnx: Done!

pmderodat (dismissed):

contrib/lkt/language/prelude.lkt
```diff
@@ -54,6 +55,10 @@ struct ASTList implements Sized, Indexable[T], Iterator[T] {
 @builtin class Equation {
 }
 
+@builtin struct PropertyError {
+    exception_message : String = "PropertyError exception"
```
thvnx, line 59: um... doesn't stick to new coding style I think?  Should be `exception_message:` without space, right?

pmderodat (approved):
That’s correct, thanks for fixing it. Until we have a style checker for Lkt, I’m afraid some errors will still fall through the cracks… One more thing to do when we’re done with the transition ;-)

PR #550, opened 2021-09-29
RA22-015: Fix dsl_unparse after recent changes.

Roldak:


langkit/dsl_unparse.py
```diff
@@ -307,6 +307,16 @@ def matches(n):
 
         return self._with_current_node(binop_node)
 
+    def is_call_to(self, names):
+        """
+        Return whether the cursor is currently on a call to a function of one
+        of the given names.
+
+        :type names: list[str]
```
pmderodat, line 315: Is there a reason not to use Python3’s type annotations here?

pmderodat (approved):
Approving as Romain is away today and we need this fix for the next stable branch bump. I’ll let Romain do updates in another PR.

PR #552, opened 2021-09-30
First step to ArrayLiteral type resolution support

thvnx:


contrib/lkt/language/parser.py
```diff
@@ -647,6 +647,26 @@ def designated_scope():
             default_val=Entity.check_expr_type.type_scope
         )
 
+    @langkit_property(return_type=T.TypeDecl.entity)
+    def array_element_type():
+        # ArrayLiteral elements type is its parent expected type's actuals
+        # type. As class Array takes only one type, actuals.at(0) is the
+        # ArrayLiteral elements type.
```
pmderodat, line 654: Can you please separate the docstring (user API doc for this property) from the implementation comment? For instance:
```python
"""
If this expression computes an array, return the type of its elements.
Return a null ``TypeDecl`` otherwise.
"""
# As the generic Array class has only one type parameter,
# ``get_actuals.at(0)`` is the element type.
```

testsuite/tests/contrib/lkt_semantic/array_type/test.lkt
```diff
@@ -0,0 +1,45 @@
+# TODO: support @invalid
```
thvnx, line 1: I guess I can remove this comment and uncomment all the `@invalid` lines below? Since it prints diagnostics we should be able to tackle regressions if any.

contrib/lkt/language/parser.py
```diff
@@ -647,6 +647,26 @@ def designated_scope():
...
+            lambda etype=Entity.expected_type.cast(T.InstantiatedGenericType):
+            Cond(
+                # No expected type found
+                etype.is_null,
+                No(T.TypeDecl.entity),
+
+                etype.get_inner_type.as_entity == Entity.array_gen_type,
+                etype.get_actuals.at_or_raise(0),
+
+                # Unreachable
```
pmderodat, line 665: Shouldn’t it be reachable, for instance when calling this property on an expression that returns a ``LexicalEnv[RootNode]``? I suppose this can happen when trying to analyze `val a: LexicalEnv[RootType] = [1]`.

testsuite/tests/contrib/lkt_semantic/array_type/test.lkt
```diff
@@ -0,0 +1,45 @@
...
+val b: Array[Int] = [1]
+val c: Array[Int] = [1, 2, 3]
+
+# @invalid val a = []
+
+# @invalid val a = [1]
+
+val i: Int = 1
+# TODO: should infer type from i
+# val a = [i, 2, 3]
```
pmderodat, line 13: It seems that a lot of commented lines are for the same feature (inter array element type inference): I suggest to move unsupported cases to a separate testcase, uncommented and with an expected failure. (commented code is too easy to forget! ;-))

Expected failures (XFAIL tests) allow us to know that a test is failing but still run it to spot when it starts to work in the future. Look for `control` entries in the testsuite’s `test.yaml` files for inspiration, and here is the doc for this mechanism: https://e3-testsuite.readthedocs.io/en/latest/api_control.html

testsuite/tests/contrib/lkt_semantic/array_type/test.lkt
```diff
@@ -0,0 +1,45 @@
+# TODO: support @invalid
+
+val a: Array[Int] = []
+val b: Array[Int] = [1]
+val c: Array[Int] = [1, 2, 3]
+
+# @invalid val a = []
```
pmderodat, line 7: It’s unclear when reading this why this is commented: don’t we already have a diagnostic for this case? (same for the next)

testsuite/tests/contrib/lkt_semantic/array_type/test.lkt
```diff
@@ -0,0 +1,45 @@
...
+
+fun bar(): String = "str1"
+# TODO: should infer type from bar()
+# val a = [foo(), "str2", "str3"]
+
+val a: Array[String] = []
+val b: Array[String] = ["1"]
+val c: Array[String] = ["1", "2", "3"]
+
+# @invalid val a: Array[Int] = ['!']
```
pmderodat, line 31: Here too I would expect an error message: can we uncomment this line?

testsuite/tests/contrib/lkt_semantic/array_type/test.lkt
```diff
@@ -0,0 +1,45 @@
...
+val b: Array[Array[Int]] = [[1]]
+val c: Array[Array[Int]] = [[1, 2, 3]]
+
+val a: Array[Array[Int]] = [[], []]
+val b: Array[Array[Int]] = [[1], [1]]
+val c: Array[Array[Int]] = [[1, 2, 3], [1, 2, 3]]
+
+val a: Array[Array[Array[Array[Int]]]] = [[[[],[]],[]],[]]
+
+# @invalid val a: Array[Array[Int]] = [1]
```
pmderodat, line 44: Here too

pmderodat (changes requested):
Thanks for your work, Laurent!

Whenever it is possible, we try to put tests for a feature/bugfix in the same commit as the implementation: can you merge the two commits?

Also, can you add the TN to the commit message, for tracking purposes? Thanks!

testsuite/tests/contrib/lkt_semantic/array_type/test.lkt
thvnx: > Thanks for your work, Laurent!
> 
> Whenever it is possible, we try to put tests for a feature/bugfix in the same commit as the implementation: can you merge the two commits?
> 
> Also, can you add the TN to the commit message, for tracking purposes? Thanks!

You're welcome ;). Sure I'll do that!

contrib/lkt/language/parser.py
thvnx: you're right!

testsuite/tests/contrib/lkt_semantic/array_type_concrete/test.out
```diff
@@ -0,0 +1,96 @@
+Id   <RefId "Int" test.lkt:1:8-1:11>
```
pmderodat, line 1: I believe this file contains “garbade” (i.e. does not correspond to the actual test). I suggest make it an empty file to avoid confusion.

testsuite/tests/contrib/lkt_semantic/array_type_concrete/test.lkt
```diff
@@ -0,0 +1,15 @@
+val i: Int = 1
+# TODO: should infer type from i
+val a = [i, 2, 3]
+
+val s: String = "str1"
+# TODO: should infer type from s
+val a = [str, "str2"]
```
pmderodat, line 7: Typo: should be `s` (not `str`)

testsuite/tests/contrib/lkt_semantic/array_type_concrete/test.lkt
```diff
@@ -0,0 +1,15 @@
...
+
+fun foo(): Int = 1
+# TODO: should infer type from foo()
+val a = [foo(), 2, 3]
+
+fun bar(): String = "str1"
+# TODO: should infer type from bar()
+val a = [foo(), "str2", "str3"]
```
pmderodat, line 15: Pasto: should be `bar()`

Also, I think it would be valuable to include an invalid case that mixes type inference and inconsistent types. For instance:

```
val a4 = [foo(), bar()]
```

testsuite/tests/contrib/lkt_semantic/array_type_concrete/test.lkt
```diff
@@ -0,0 +1,15 @@
+val i: Int = 1
+# TODO: should infer type from i
+val a = [i, 2, 3]
```
pmderodat, line 3: All array values in this file are called `a`, which should probably be invalid in the future. Could you rename them? For intsance a1, a2, …

testsuite/tests/contrib/lkt_semantic/array_type/test.lkt
pmderodat: Ah indeed. Yes, even though `@invalid` is ignored, it looks like the test driver still has the behavior we want (print the error when there is one), so it’s ok.

testsuite/tests/contrib/lkt_semantic/array_type_expected/test.lkt
```diff
@@ -0,0 +1,25 @@
+val a: Array[Int] = []
+val b: Array[Int] = [1]
+val c: Array[Int] = [1, 2, 3]
+
+@invalid val a = []
```
pmderodat: For clarity, I suggest adding comments to explain why invalid declarations are… invalid ^^

Here, for instance:
```
# It is impossible to infer the element type of an empty array literal, without
# the help of type info from the context
@invalid val a = []
```
Also here too, please avoid using the same identifier for two entities in the same scope.

testsuite/tests/contrib/lkt_semantic/array_type_concrete/test.out
thvnx: This is the output of lkt_toolbox on the test.lkt file.

testsuite/tests/contrib/lkt_semantic/array_type_expected/test.out
```diff
@@ -0,0 +1,314 @@
...
+     references <GenericDecl prelude: "Array">
+
+Id   <RefId "Array" test.lkt:25:23-25:28>
+     references <GenericDecl prelude: "Array">
+
+Id   <RefId "Int" test.lkt:25:29-25:32>
+     references <StructDecl prelude: "Int">
+
+test.lkt:25:40: error: ambiguous type for expression
+24 | @invalid val a: Array[Array[Int]] = [[[1]]]
```
pmderodat: This error message is surprising: I would have expected rather something like:
```
Mismatched types: expected `Int`, got `Array[Int]` instead
```
@raph-amiard what do you think?

testsuite/tests/contrib/lkt_semantic/array_type_concrete/test.lkt
thvnx: Correct, and the test.out below tells me that also, sorry to missed that.

testsuite/tests/contrib/lkt_semantic/array_type_concrete/test.lkt
thvnx: Will be done

testsuite/tests/contrib/lkt_semantic/array_type_concrete/test.lkt
thvnx: Sure!

pmderodat (changes requested):
Thanks a lot for the update. The new XFAIL test has minor issues, and I’m wondering about an error case in the working test.

Also, in the commit message:
> this patch allows to type well typed arrays (i.e., a generic_type has been declared for them)

This is partially wrong (not all well typed arrays *literals* are supported right now), could make one think that invalid arrays literals are not supported (they are supposed to, in the sense that we emit a good error message for them) and I’m not sure what you mean about the `generic_type`. Correct me if I’m wrong or forgot something, but what about:
> This patch implements the typing of array literals when the information for element type comes from the context.

Finally:
> Nevertheless, the parser is not able yet to infer its type from its context-free type elements.

“… is not able yet to infer the type of array literals from the context-free type of their elements”
(it’s array literals that we support here: we already support some expressions that deal with arrays)

contrib/lkt/language/parser.py
```diff
@@ -647,6 +647,29 @@ def designated_scope():
...
+    @langkit_property(return_type=T.TypeDecl.entity)
+    def array_element_type():
+        """
+        If this expression computes an array, return the type of its elements.
+        Return a null ``TypeDecl`` otherwise.
+        """
+        # As the generic Array class has only one type parameter,
+        # ``get_actuals.at(0)`` is the element type.
+        return Let(
+            lambda etype=Entity.expected_type.cast(T.InstantiatedGenericType):
```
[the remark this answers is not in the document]
thvnx: How can I use the `etype` in that case?

Something like:
```
Entity.expected_type.cast(T.InstantiatedGenericType).then(
  lambda etype: Cond(
    etype == condtype,
    etype.get...

    raise error
  ) 
  default_val=No(T.Type.entity)
)
```

contrib/lkt/language/parser.py
```diff
@@ -647,6 +647,29 @@ def designated_scope():
...
+            Cond(
+                # No expected type found
+                etype.is_null,
+                No(T.TypeDecl.entity),
+
+                etype.get_inner_type.as_entity == Entity.array_gen_type,
+                etype.get_actuals.at_or_raise(0),
+
+                # TODO: add unsupported cases or emit proper diagnostic
```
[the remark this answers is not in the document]
thvnx: IIRC, @pmderodat was thinking of something like: `LexicalEnv[Node] = [1]`.

testsuite/tests/contrib/lkt_semantic/array_type_expected/test.out
thvnx: Shouldn't be something like:

```Mismatched types: expected `Array[Array[Array[Int]]]`, got `Array[Array[Int]]` instead```

?

Comment (thvnx) [reply testsuite/tests/contrib/lkt_semantic/array_type_expected/test.out]: Actually, looks good to me as `[[[1]]]` type cannot be inferred. At the moment, we can only emit this kind of diagnostic in such cases.

File: contrib/lkt/language/parser.py (hunk 15 above)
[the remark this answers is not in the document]
Comment (thvnx) [reply contrib/lkt/language/parser.py:668]: ping @raph-amiard, should I reword the comment?

File: testsuite/tests/contrib/lkt_semantic/array_type_concrete/test.out (hunk 8 above)
Comment (pmderodat) [reply testsuite/tests/contrib/lkt_semantic/array_type_concrete/test.out:1]: Yes, but that’s not the expected output: for instance nodes from the prelude have slocs (they shouln’t in baselines), line numbers are slightly off and most importantly, this file makes it feel like the typing errors are expected (because of the lack of type inference from array literal elements). And if you update the test.out to be the current output (with the `Resolving test.lkt` header), you’ll even get an UOK (unexpected success) test instead of an XFAIL one (expected failure), making one think that the test actually passes, whereas it should be failing (waiting for the TODO to be implemented).

File: testsuite/tests/contrib/lkt_semantic/array_type_expected/test.lkt (hunk 12 above)
Comment (pmderodat) [reply testsuite/tests/contrib/lkt_semantic/array_type_expected/test.lkt]: Nice, thanks!

File: testsuite/tests/contrib/lkt_semantic/array_type_expected/test.out (hunk 13 above)
Comment (pmderodat) [reply testsuite/tests/contrib/lkt_semantic/array_type_expected/test.out]: I still would like to have @raph-amiard’s opinion on this

File: testsuite/tests/contrib/lkt_semantic/array_type_concrete/test.out (hunk 8 above)
Comment (thvnx) [reply testsuite/tests/contrib/lkt_semantic/array_type_concrete/test.out:1]: OK, I'll empty this file then!

File: testsuite/tests/contrib/lkt_semantic/array_type_expected/test.out (hunk 13 above)
Comment (pmderodat) [reply testsuite/tests/contrib/lkt_semantic/array_type_expected/test.out]: Oh I forgot to answer you:

We can’t infer the type of `[[[1]]]` without help from the context, so we can’t say that we got `Array[Array[Array[Int]]]`. What I would have thought we could do would have been:

* Try to match `Array[Array[Int]]` with `[[[1]]]`: both are arrays, so…
* Try to match `Array[Int]` with `[[1]]`: both are arrays, so…
* Try to match `Int` with `[1]`: one is an array and not the former, so we can emit `expected Int, got Array[Int] instead`

File: contrib/lkt/language/parser.py (hunk 15 above)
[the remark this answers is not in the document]
Comment (pmderodat) [reply contrib/lkt/language/parser.py]: That’s indeed a case I thought it was worth checking: can you extend a testcase with it? For now it seems that for it, we get the usual error message:

```
error: ambiguous type for expression
@invalid val foo: LexicalEnv[Int] = [1]
```

Which is probably okay for now, but as discussed live we will need a better (more helpful) error message in the future. I don’t know if this is the correct place to implement it though, so I don’t know if this comment should be here. @raph-amiard what do you think?

File: testsuite/tests/properties/character/test.py [hunk 16]
```diff
@@ -37,5 +37,8 @@ def text_identity(s=T.String):
```
pmderodat, line 37: Please move this commit before 458e6b23454297f4db798b254adc445ef956d184, otherwise the test will fail on that commit.

contrib/lkt/language/parser.py
```diff
@@ -2208,6 +2208,41 @@ def expected_type_predicate(expected_type=T.TypeDecl.entity):
     def invalid_expected_type_error_name():
```
pmderodat, line 2219: (for the commit message)

as weel -> as well

from the array elements known types -> from the context-free types for array elements
(or even better: just “for elements”, as we know we’re talking about arrays)

Therefore*
(see https://forum.wordreference.com/threads/therefor-and-therefore.1261337/)

contrib/lkt/language/parser.py
```diff
@@ -2208,6 +2208,41 @@ def expected_type_predicate(expected_type=T.TypeDecl.entity):
...
         return S("an array")
 
+    @langkit_property()
+    def expr_context_free_type():
+        """
+        Infer type from array elements context_free_type.
```
pmderodat, line 2225: Try to infer the array type from elements's context-free types*

(type inference is just tentative, and let’s use plain English when possible for the high-level explanation in the docstring)

contrib/lkt/language/parser.py
```diff
@@ -2208,6 +2208,41 @@ def expected_type_predicate(expected_type=T.TypeDecl.entity):
...
+        """
+        # Get the context_free_type of the array elements if any
+        types = Var(
+            Entity.exprs.filter(
+                # Filter out Lit expressions to avoid infinite recursion
+                lambda e: Not(
+                    Or(e.is_a(T.NumLit),
+                       e.is_a(T.CharLit),
+                       e.is_a(T.StringLit))
```
pmderodat, line 2234: `.is_a()` accepts multiple type arguments, so this can be simplified to: `Not(e.is_a(T.NumLite, T.CharLit, T.StringLit))`

contrib/lkt/language/parser.py
```diff
@@ -2208,6 +2208,41 @@ def expected_type_predicate(expected_type=T.TypeDecl.entity):
...
+                )
+            ).filtermap(
+                # Keep only elements context_free_types
```
pmderodat, line 2237: Keep non-null context-free types only*

But thinking more about this, I think this is an architecture smell: first, we need to keep here the list of nodes for which `expr_context_free` may end up calling back `ArrayLiteral`’s, it’s a code maintainance issue.

But maybe more importantly, it feels wrong that `Lit.expr_context_free_type` does checks the context (i.e. info coming from its parents) in order to compute the type. What about moving the code currently in `Lit.expr_context_free_type_impl` (which checks the context) to `Expr.array_element_type`? (as this one already checks the context, through its call to `.expected_type`) This way we can get rid of the above type-based filtering and preserve the meaning of “context-free type”.

contrib/lkt/language/parser.py
```diff
@@ -2208,6 +2208,41 @@ def expected_type_predicate(expected_type=T.TypeDecl.entity):
...
+                lambda e: e.expr_context_free_type.node,
```
pmderodat, line 2238: The `.node` part looks wrong: for instance with a `LexicalEnv[T]` type, it will remove the instantiation information, so we’ll only have `LexicalEnv`.

contrib/lkt/language/parser.py
```diff
@@ -2208,6 +2208,41 @@ def expected_type_predicate(expected_type=T.TypeDecl.entity):
...
+                lambda e: Not(e.expr_context_free_type.is_null)
+            )
+        )
+
+        has_cf_type = Var(
+            If(types.length == 0,
+               False,
+               types.all(lambda t: t == types.at(0)))
```
pmderodat, line 2246: This can be simplified to: `Not(types.length == 0) & types.all(lambda t: t == types.at(0))`

contrib/lkt/language/parser.py
```diff
@@ -2208,6 +2208,41 @@ def expected_type_predicate(expected_type=T.TypeDecl.entity):
...
+        )
+
+        # Return the type of the ArrayLiteral: an Array of elememts
+        # context_free_types if they are identical, no type otherwise.
+        return If(has_cf_type,
+                  Self.array_gen_type.get_instantiated_type(
+                      [types.at(0)]
+                  ),
```
pmderodat, line 2254: I think we should create a property to instantiate Array, as it look very useful (and will make the code more readable):

```
If(
    has_cf_type,
    Self.get_array_type(types.at(0)),
    No(T.TypeDecl)
).as_entity

File: contrib/lkt/language/parser.py [hunk 24]
```diff
@@ -2199,6 +2199,17 @@ class ArrayLiteral(Expr):
     """
     exprs = Field(type=T.Expr.list)
 
+    @langkit_property(return_type=T.SemanticResult.array)
+    def check_correctness_pre():
```
pmderodat, line 2203: The work here is not obvious: can you add comments to explain?

contrib/lkt/language/parser.py
```diff
@@ -2156,6 +2182,15 @@ class ArrayLiteral(Expr):
     """
     exprs = Field(type=T.Expr.list)
 
+    @langkit_property()
+    def expected_type_predicate(expected_type=T.TypeDecl.entity):
+        # Any ArrayLiteral should be an InstantiatedGenricType
```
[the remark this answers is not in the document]
thvnx: Yes, you're right. I was just working on that!

contrib/lkt/language/parser.py
thvnx: I agree, I was not feeling 100% sure of me on that point. Will rework this part.

contrib/lkt/language/parser.py
```diff
@@ -662,11 +662,9 @@ def array_element_type():
         # As the generic Array class has only one type parameter,
```
pmderodat, line 662: (in the commit message): inferred* (not infered)

contrib/lkt/language/parser.py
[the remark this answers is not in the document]
thvnx: I'll merge the two commits for more consistency here!

contrib/lkt/language/parser.py
```diff
@@ -2187,13 +2196,51 @@ class ArrayLiteral(Expr):
...
     def invalid_expected_type_error_name():
         return S("an array")
 
+    @langkit_property()
+    def expr_context_free_type():
+        # Get all non-null context-free type of exprs list
+        expected_exprs_type = Var(Self.exprs.filtermap(
+            lambda e: e.as_entity.expr_context_free_type,
+            lambda e: Not(e.as_entity.expr_context_free_type.is_null)
+        ))
```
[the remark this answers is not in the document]
thvnx: Code has been factorised.

contrib/lkt/language/parser.py
```diff
@@ -2187,13 +2196,51 @@ class ArrayLiteral(Expr):
...
+        # Check that all elements in exprs, element type is either the first
+        # context-free type given by expected_exprs_type, or a compatible one.
```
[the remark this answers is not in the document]
thvnx: Won't fix this in this PR.

PR #553, opened 2021-09-30
U928-029: enable lkt_semantic_checks for memoized_property_error test

thvnx:
Cannot reproduce the error, the test doesn't fail anymore.


pmderodat (changes requested):
Same objection as in #551

PR #554, opened 2021-10-06
U930-039: second step to ArrayLiteral type resolution support

thvnx:
Infer ArrayLiteral elements types from their already known elements's context_free_types.

Note that this PR shouldn't be based on `master` but `array-type-1` (#552).


thvnx:
merged in #552

PR #557, opened 2021-10-08
Properties: make String an opaque type 

pmderodat:


pmderodat:
@danielmercier @yakobowski Hello! Could you please review the OCaml part of the “Properties: make String an opaque type” commit? You can see this part as mostly a cleanup (no special cases for `Character.array` anymore). Thanks!

PR #576, opened 2021-11-08
Lkt lowering: add support for even more constructs

pmderodat:


langkit/lkt_lowering.py
```diff
@@ -1360,35 +1366,81 @@ def lower_base_field(
...
+                'public': annotations.export,
+                'return_type': field_type,
+                'kind': (AbstractKind.abstract
+                         if annotations.abstract
+                         else AbstractKind.concrete),
+            }
+
+        elif annotations.parse_field:
+            assert decl.f_default_val is None
+            check_source_language(
```
raph-amiard, line 1414: I'm a bit uncomfortable with those checks here: We have most of the legality checks in `parser.py`, wouldn't it be better to generally put those there ?

langkit/lkt_lowering.py
```diff
@@ -1407,6 +1410,10 @@ def lower_base_field(
                 not annotations.lazy,
                 'Parse fields cannot be lazy'
             )
+            check_source_language(
```
raph-amiard, line 1422: Same comment as before

langkit/lkt_lowering.py
```diff
@@ -1559,6 +1565,59 @@ def lower(expr: L.Expr) -> AbstractExpression:
...
+                vars = []
+                var_exprs = []
+
+                for v in expr.f_val_defs:
+                    if isinstance(v, L.ValDecl):
+                        source_name = names.Name.from_lower(v.f_syn_name.text)
+
+                        # Make sure the name used for code generation is a
+                        # valid Ada identifier.
+                        code_name = source_name.lower
```
raph-amiard, line 1600: Having this here kind of makes me uncomfortable, shouldn't this be factorized in an independent abstraction ?

testsuite/tests/properties/abstract_rt/expected_concrete_syntax.lkt
```diff
@@ -0,0 +1,19 @@
+import lexer_example
+@with_lexer(foo_lexer)
+grammar foo_grammar {
+    @main_rule main_rule <- list*(or(Example("example") | Error("error")))
+
+}
+
+@abstract class FooNode implements Node[FooNode] {
+
+    @export @not_implemented fun prop(): Int
```
raph-amiard, line 10: Not fond of the name, but I don't have anything better right now...

raph-amiard: I wonder if we shouldn't drop this feature considering the alternative:

```
    @export fun prop(): Int = raise PropertyError("prop not implemented on" & Self.image)
```

PM, @Roldak, wdyt ?

raph-amiard (changes requested):

testsuite/tests/properties/abstract_rt/expected_concrete_syntax.lkt
pmderodat: I’m completely ok to drop this feature: it probably made sense back when we didn’t have raise expressions, but I think we could get rid of it nowadays.

(otherwise I would be ok to rename it, but in another commit given that this PR does not introduce it)

langkit/lkt_lowering.py
pmderodat: As discussed live: I’ll turn these into asserts with TODOs to add the corresponding checks in liblktlang. However I would prefer to implement these TODOs in another PR.

langkit/lkt_lowering.py
pmderodat: This particular area is full of similar existing checks: I would prefer to transfer them to Liblktlang in one pass rather that just do it now for this one. Is it ok for you?

pmderodat: As discussed live: I’ll take care of porting existing checks in a later PR (this is my next job wrt Lkt)

langkit/lkt_lowering.py
pmderodat: Good idea, done!

testsuite/tests/properties/abstract_rt/expected_concrete_syntax.lkt
pmderodat: As discussed live, I’ve removed @not_implemented in dsl_unparse (nothing to do left in Lkt in that domain)

raph-amiard (approved):

PR #578, opened 2021-11-10
Lkt: check for @invalid annotations in lkt_toolbox

thvnx:
lkt_toolbox now raise an error if a declaration annotated with
@invalid doesn't emit any diagnostic, and conversely, it also checks
that a diagnostic is related to an @invalid declaration.

TN: UB03-005

### DIFF
--- a/contrib/lkt/extensions/mains/lkt_toolbox.adb
+++ b/contrib/lkt/extensions/mains/lkt_toolbox.adb
@@ -1,3 +1,5 @@
+with Ada.Command_Line;            use Ada.Command_Line;
+with Ada.Containers.Hashed_Maps;
 with Ada.Directories;             use Ada.Directories;
 with Ada.Exceptions;
 with Ada.Strings.Fixed;           use Ada.Strings.Fixed;
@@ -13,6 +15,7 @@ with Langkit_Support.Diagnostics; use Langkit_Support.Diagnostics;
 with Langkit_Support.Diagnostics.Output;
 use Langkit_Support.Diagnostics.Output;
 
+with Langkit_Support.Slocs;       use Langkit_Support.Slocs;
 with Langkit_Support.Text;        use Langkit_Support.Text;
 
 with Liblktlang.Analysis;         use Liblktlang.Analysis;
@@ -42,12 +45,36 @@ procedure Lkt_Toolbox is
 
    use Liblktlang;
 
+   package Invalid_Decl_Maps is new Ada.Containers.Hashed_Maps
+     (Key_Type        => Analysis.LK_Node,
+      Element_Type    => Boolean,
+      Hash            => Analysis.Hash,
+      Equivalent_Keys => "=");
+
    procedure Print_Semantic_Result
      (S : Analysis.Semantic_Result; Unit : Analysis.Analysis_Unit);
    --  Print a semantic result
 
    function Format_Node (Decl_Node : Decl'Class) return String;
    --  Format node for semantic result printing
+
+   procedure Print_Lkt_Toolbox_Diagnostic
+     (Node : LK_Node'Class; Message : Wide_Wide_String);
+   --  Internal wrapper to ``Print_Diagnostic`` used by lkt_toolbox to print
+   --  additional diagnostics.
+
+   function Populate_Invalid_Decl_Map
+     (Node : Analysis.LK_Node'Class) return Common.Visit_Status;
+   --  Populate ``Invalid_Decl_Map`` and reject declarations with ``@invalid``
+   --  annotations that are nested in another declaration annotated with
+   --  ``@invalid``.
+
+   Invalid_Decl_Map : Invalid_Decl_Maps.Map;
+   --  Map of declarations annotated with ``@invalid``. The boolean elements of
+   --  the map are initialized to ``False`` and set to ``True`` whenever a
+   --  diagnostic is emitted for the related declaration. Therefore, this map
+   --  is used to check that at least one diagnostic has been emitted for each
+   --  declaration annotated with ``@invalid``.
 
    -----------------
    -- Format_Node --
@@ -60,6 +87,20 @@ procedure Lkt_Toolbox is
       return Decl_Node.P_As_Bare_Decl.Image;
    end Format_Node;
 
+   ----------------------------------
+   -- Print_Lkt_Toolbox_Diagnostic --
+   ----------------------------------
+
+   procedure Print_Lkt_Toolbox_Diagnostic
+     (Node : LK_Node'Class; Message : Wide_Wide_String)
+   is
+      Sloc_Range : constant Source_Location_Range := Node.Sloc_Range;
+      Unit       : constant Analysis.Analysis_Unit := Node.Unit;
+      Path       : constant String := Simple_Name (Unit.Get_Filename);
+   begin
+      Print_Diagnostic ((Sloc_Range, To_Unbounded_Text (Message)), Unit, Path);
+   end Print_Lkt_Toolbox_Diagnostic;
+
    ---------------------------
    -- Print_Semantic_Result --
    ---------------------------
@@ -67,33 +108,77 @@ procedure Lkt_Toolbox is
    procedure Print_Semantic_Result
      (S : Analysis.Semantic_Result; Unit : Analysis.Analysis_Unit)
    is
+      Node : constant LK_Node'Class := Analysis.Node (S);
    begin
       if Analysis.Error_Message (S) /= "" then
          declare
             Diag : constant Diagnostic :=
-              (Analysis.Node (S).Sloc_Range,
+              (Node.Sloc_Range,
                To_Unbounded_Text (Analysis.Error_Message (S)));
          begin
+            --  Emit an error if the declaration including ``Node`` has no
+            --  ``@invalid`` annotation. Update ``Invalid_Decl_Map`` otherwise.
+
+            if Node.P_Topmost_Invalid_Decl.Is_Null then
+               Set_Exit_Status (1);
+
+               Print_Lkt_Toolbox_Diagnostic
+                 (Node,
+                  "unexpected diagnostic, is @invalid annotation missing?");
+            else
+               Invalid_Decl_Map (Node.P_Topmost_Invalid_Decl) := True;
+            end if;
+
             Print_Diagnostic
-              (Diag, Unit,
-               Simple_Name (Analysis.Node (S).Unit.Get_Filename));
+              (Diag, Unit, Simple_Name (Node.Unit.Get_Filename));
          end;
       elsif
         not Arg.Check_Only.Get
         and then not Analysis.Result_Type (S).Is_Null
       then
-         Put_Line ("Expr " & Analysis.Node (S).Image);
+         Put_Line ("Expr " & Node.Image);
          Put_Line ("     has type " & Analysis.Result_Type (S).Image);
          New_Line;
       elsif
       not Arg.Check_Only.Get
         and then not Analysis.Result_Ref (S).Is_Null
       then
-         Put_Line ("Id   " & Analysis.Node (S).Image);
+         Put_Line ("Id   " & Node.Image);
          Put_Line ("     references " & Format_Node (Analysis.Result_Ref (S)));
          New_Line;
       end if;
    end Print_Semantic_Result;
+
+   -------------------------------
+   -- Populate_Invalid_Decl_Map --
+   -------------------------------
+
+   function Populate_Invalid_Decl_Map
+     (Node : Analysis.LK_Node'Class) return Common.Visit_Status
+   is
+      use type Common.LK_Node_Kind_Type;
+   begin
+      --  Populate ``Invalid_Decl_Map`` with declarations annotated with
+      --  ``@invalid``.
+
+      if Node.Kind = Common.Lkt_Full_Decl
+         and then Node.As_Full_Decl.P_Has_Annotation
+           (To_Unbounded_Text ("invalid"))
+      then
+         --  ``P_Topmost_Invalid_Decl`` should return the same node. In that
+         --  case, include this node in the map, otherwise nested ``@invalid``
+         --  declarations have been detected: emit a diagnostic.
+
+         if Invalid_Decl_Map.Contains (Node.P_Topmost_Invalid_Decl) then
+            Set_Exit_Status (1);
+
+            Print_Lkt_Toolbox_Diagnostic (Node, "nested @invalid declaration");
+         else
+            Invalid_Decl_Map.Include (Node.As_LK_Node, False);
+         end if;
+      end if;
+      return Common.Into;
+   end Populate_Invalid_Decl_Map;
 
    Ctx : constant Analysis.Analysis_Context := Analysis.Create_Context;
 begin
@@ -119,6 +204,8 @@ begin
                return;
             end if;
 
+            Unit.Root.Traverse (Populate_Invalid_Decl_Map'Access);
+
             declare
                Diags : constant Analysis.Tree_Semantic_Result :=
                  Unit.Root.P_Check_Semantic;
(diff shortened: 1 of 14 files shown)


contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -42,13 +43,27 @@ procedure Lkt_Toolbox is
...
    function Format_Node (Decl_Node : Decl'Class) return String;
    --  Format node for semantic result printing
 
+   function Populate_Invalid_Decl_Map
+      (Node : Analysis.LK_Node'Class) return Common.Visit_Status;
+   --  Populate invalid declarations map and check for nested invalid
+   --  declarations.
+
+   Invalid_Decl_Map : Invalid_Decl_Maps.Map;
+   --  Map of declarations annotated with @invalid
```
pmderodat, line 65: This lacks a description of what the maps map to (i.e. the meaning of boolean there)

contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -42,13 +43,27 @@ procedure Lkt_Toolbox is
...
    procedure Print_Semantic_Result
      (S : Analysis.Semantic_Result; Unit : Analysis.Analysis_Unit);
    --  Print a semantic result
 
    function Format_Node (Decl_Node : Decl'Class) return String;
    --  Format node for semantic result printing
 
+   function Populate_Invalid_Decl_Map
+      (Node : Analysis.LK_Node'Class) return Common.Visit_Status;
+   --  Populate invalid declarations map and check for nested invalid
```
pmderodat, line 61: I would suggest to directly mention the map variable that this populates, and be more specific “invalid declaration” is a bit confusing: a decl with the @invalid annotation versus a declaration that is actually invalid: we can’t assume there that they are the same, otherwise lkt_toolbox would work only on correct tests, while the point here is to check test correctness.
```
Populate ``Invalid_Decl_Map`` and reject declarations with ``@invalid``
annotations that are nested in another declaration annotated with ``@invalid``.
```

contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -77,6 +92,17 @@ procedure Lkt_Toolbox is
             Print_Diagnostic
               (Diag, Unit,
                Simple_Name (Analysis.Node (S).Unit.Get_Filename));
+
+            --  Raise an error if the declaration including ``Node (S)`` has no
+            --  ``@invalid`` annotation. Update ``Invalid_Decl_Map`` otherwise.
+
+            if Analysis.Node (S).P_Topmost_Invalid_Decl.Is_Null
```
pmderodat, line 99: Please refactor to avoid the `Analysis.Node (S)` repetitions

contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -95,6 +121,36 @@ procedure Lkt_Toolbox is
       end if;
    end Print_Semantic_Result;
 
+   -------------------------------
+   -- Populate_Invalid_Decl_Map --
+   -------------------------------
+
+   function Populate_Invalid_Decl_Map
+      (Node : Analysis.LK_Node'Class) return Common.Visit_Status is
```
pmderodat, line 129: We generally put `is` on its own line when there are declarations before the `begin` keyword (less confusing layout)

contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -42,13 +43,27 @@ procedure Lkt_Toolbox is
...
+
    procedure Print_Semantic_Result
      (S : Analysis.Semantic_Result; Unit : Analysis.Analysis_Unit);
    --  Print a semantic result
 
    function Format_Node (Decl_Node : Decl'Class) return String;
    --  Format node for semantic result printing
 
+   function Populate_Invalid_Decl_Map
+      (Node : Analysis.LK_Node'Class) return Common.Visit_Status;
```
pmderodat, line 60: There is one extra indentation space: as discussed live earlier, we add only 2 columns when the next line is an opening paren. (same multiple times below)

contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -95,6 +121,36 @@ procedure Lkt_Toolbox is
...
+      if Node.Kind = Common.Lkt_Full_Decl
+         and then Node.As_Full_Decl.P_Has_Annotation
+            (To_Unbounded_Text ("invalid"))
+      then
+         --  ``P_Topmost_Invalid_Decl`` should return the same node. In that
+         --  case, include this node in the map, otherwise nested ``@invalid``
+         --  declaration has been detected: raise an error.
+
+         if Invalid_Decl_Map.Contains
+               (Node.P_Topmost_Invalid_Decl.As_LK_Node)
```
pmderodat, line 144: The other map operation above does not need the `As_Lk_Node` conversion: why is it needed here? (same below)

contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -127,6 +185,17 @@ begin
                   Print_Semantic_Result (D, Unit);
                end loop;
             end;
+
+            --  Ensure that all ``@invalid`` declarations in the map have been
+            --  reported.
+
+            for E in Invalid_Decl_Map.Iterate loop
+               if not Invalid_Decl_Maps.Element (E)
+               then
```
pmderodat, line 194: Please move `then` on the line above

contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -127,6 +185,17 @@ begin
...
+                  raise Constraint_Error;
```
pmderodat, line 195: Same remark as above wrt. the exception

contrib/lkt/language/parser.py
```diff
@@ -470,6 +470,28 @@ def first_no_paren_parent():
             Entity.parent
         )
 
+    @langkit_property(public=True, return_type=T.LKNode.entity)
+    def topmost_invalid_decl(
+        topmost=(T.LKNode.entity, No(T.LKNode.entity))
+    ):
+        """
+        Return the topmost FullDecl annotated with ``@invalid``,
+        null otherwise.
```
pmderodat, line 479: Can you document what the `topmost` argument is? Also clarify whether the current node is included in the search (it looks like it’s not looking at the implementation, but that should be documented here).

pmderodat (changes requested):

contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -77,6 +92,17 @@ procedure Lkt_Toolbox is
             Print_Diagnostic
               (Diag, Unit,
                Simple_Name (Analysis.Node (S).Unit.Get_Filename));
+
+            --  Raise an error if the declaration including ``Node (S)`` has no
+            --  ``@invalid`` annotation. Update ``Invalid_Decl_Map`` otherwise.
+
+            if Analysis.Node (S).P_Topmost_Invalid_Decl.Is_Null
+            then
+               raise Constraint_Error;
```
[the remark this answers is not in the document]
thvnx: Should be good now but I couldn't add some tests for those new errors since lkt_toolbox returns a non-zero status code.

Should I add some? This will require to add a new lkt_toolbox option to set the status code, and to modify lkt_driver to manage arguments.

contrib/lkt/extensions/mains/lkt_toolbox.adb
pmderodat: As discussed live the other day, you can modify `Node`'s declaration to be a simple `LK_Node`, and initialize it with `Analysis.Node (S).As_LK_Node`.

pmderodat (changes requested):

contrib/lkt/extensions/mains/lkt_toolbox.adb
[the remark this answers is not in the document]
pmderodat: I think it’s fine not to add tests, but it’s for sure best to have some: your call!

contrib/lkt/extensions/mains/lkt_toolbox.adb
thvnx: I still end up with a compilation error when trying to get rid of this `As_LK_Node` conversion [here](https://github.com/AdaCore/langkit/blob/f87bb05f218fc176d8986d872e72c2238886f164/contrib/lkt/extensions/mains/lkt_toolbox.adb#L177). As it is used in the traversal function, `Node` should be of type `LK_Node'Class` no? I guess we should discuss it again.

contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -67,34 +88,83 @@ procedure Lkt_Toolbox is
...
+            --  Emit an error if the declaration including ``Node`` has no
+            --  ``@invalid`` annotation. Update ``Invalid_Decl_Map`` otherwise.
+
+            if Node.P_Topmost_Invalid_Decl.Is_Null
+            then
+               Set_Exit_Status (1);
+
+               Print_Diagnostic ((Node.Sloc_Range, To_Unbounded_Text
+                 ("unexpected diagnostic, is @invalid annotation missing?")),
+                  Unit, Simple_Name (Node.Unit.Get_Filename));
```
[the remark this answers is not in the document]
thvnx: I've added a new procedure to take care of that. Thanks for the suggestion.

contrib/lkt/language/parser.py
```diff
@@ -470,6 +470,31 @@ def first_no_paren_parent():
...
+    @langkit_property(public=True, return_type=T.LKNode.entity)
+    def topmost_invalid_decl(
+        topmost=(T.LKNode.entity, No(T.LKNode.entity))
+    ):
+        """
+        Return the ``topmost`` FullDecl annotated with ``@invalid``,
+        null otherwise. ``topmost`` argument is the topmost declaration
+        annotated with ``@invalid`` from the current node to the root node if
+        it exists, null otherwise. It is used to get the first ``@invalid``
+        declaration in case where they are nested.
```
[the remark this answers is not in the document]
thvnx: Actually, it does start from `Entity` (and therefore can return it): it starts by looking at the parent in order to start the recursion from it, by setting `topmost` to `Entity` if it is an `@invalid` declaration. There is only one case where it cannot return `Entity`: when it has no parent, i.e. `Entity` is the root node, which is OK since it cannot be a declaration annotated with `@invalid`.
```
        return Entity.parent.then(
            lambda p: p.topmost_invalid_decl(
                Entity.is_a(FullDecl).then(
                    lambda _: Entity.cast(FullDecl)
                    .has_annotation('invalid').then(
                        lambda _: Entity,             # <- starts from Entity
                        default_val=topmost
                    ),
                    default_val=topmost
                )
            ),
            default_val=topmost
        )
```

contrib/lkt/extensions/mains/lkt_toolbox.adb
pmderodat: That’s correct: Node being a `LK_Node'Class` value because of the `Traverse` subprogram (and you can’t change that), you need to call `.As_LK_Node` to pass it as a `LK_Node` value.

It seems that GitHub won’t let me see the original commit for which I wrote “same below”, so I can’t double check: sorry if I was misleading. The simplification was valid for `Node.P_Topmost_Invalid_Decl` in any case.

contrib/lkt/language/parser.py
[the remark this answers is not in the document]
pmderodat: Oh I see, you are correct, thanks. I find this specific recursion hard to proof-read, there are odd repetitions (`default_val=topmost` appears 3 times) and the fact that the topmost argument (which is there only as an implementation detail) leaks to the user API is a smell. I would suggest to use `.parents()` in order to switch to an iterative (and straightforward) form. Something such as:

```python
return Self.parents().filter(
    lambda p:
    p.cast(FullDecl).then(lambda fd: fd.has_annotation("invalid"))
).at(-1)
```

contrib/lkt/extensions/mains/lkt_toolbox.adb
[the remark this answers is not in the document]
thvnx: Will keep that for later then!

contrib/lkt/language/parser.py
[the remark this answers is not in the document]
thvnx: Done!

contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -60,41 +87,99 @@ procedure Lkt_Toolbox is
...
+      --  Populate ``Invalid_Decl_Map`` with declarations annotated with
+      --  ``@invalid``.
+
+      if Node.Kind = Common.Lkt_Full_Decl
+         and then Node.As_Full_Decl.P_Has_Annotation
+           (To_Unbounded_Text ("invalid"))
+      then
+         --  ``P_Topmost_Invalid_Decl`` should return the same node. In that
+         --  case, include this node in the map, otherwise nested ``@invalid``
+         --  declaration has been detected: emit a diagnostic.
```
pmderodat, line 169: declarations*

contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -42,13 +45,37 @@ procedure Lkt_Toolbox is
...
+
    procedure Print_Semantic_Result
      (S : Analysis.Semantic_Result; Unit : Analysis.Analysis_Unit);
    --  Print a semantic result
 
    function Format_Node (Decl_Node : Decl'Class) return String;
    --  Format node for semantic result printing
 
+   procedure Print_Lkt_Toolbox_Diagnostic
+     (Node : LK_Node'Class; Message : Unbounded_Text_Type);
```
pmderodat, line 62: It looks like all calls to this procedure use `To_Unbounded_Text`: what about making it part of the factorization too?

contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -60,41 +87,99 @@ procedure Lkt_Toolbox is
...
+              (Node.Sloc_Range,
                To_Unbounded_Text (Analysis.Error_Message (S)));
          begin
+            --  Emit an error if the declaration including ``Node`` has no
+            --  ``@invalid`` annotation. Update ``Invalid_Decl_Map`` otherwise.
+
+            if Node.P_Topmost_Invalid_Decl.Is_Null then
+               Set_Exit_Status (1);
+
+               Print_Lkt_Toolbox_Diagnostic (Node, To_Unbounded_Text
```
pmderodat, line 125: This will probably get refactored, but: this still has the source layout issue I mentionned in https://github.com/AdaCore/langkit/pull/578#discussion_r748436038

pmderodat (changes requested):
LGTM after minor nits. Thanks!

contrib/lkt/extensions/mains/lkt_toolbox.adb
thvnx: Done!

contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -60,41 +87,98 @@ procedure Lkt_Toolbox is
...
                To_Unbounded_Text (Analysis.Error_Message (S)));
          begin
+            --  Emit an error if the declaration including ``Node`` has no
+            --  ``@invalid`` annotation. Update ``Invalid_Decl_Map`` otherwise.
+
+            if Node.P_Topmost_Invalid_Decl.Is_Null then
+               Set_Exit_Status (1);
+
+               Print_Lkt_Toolbox_Diagnostic (Node,
+                  "unexpected diagnostic, is @invalid annotation missing?");
```
pmderodat, line 126: Not quite: should be

```
Print_Lkt_Toolbox_Diagnostic
  (Node, "unexpected diagnostic, is @invalid annotation missing?");
```
In general: paren subexpressions should not go backwards in column numbers compared to the opening paren.

thvnx: Thanks! Hope it's good now!

pmderodat: Yes, thanks!

pmderodat (approved):

PR #581, opened 2021-11-10
Use collections.abc.Sequence as the alias is set to be removed

### LINKED ISSUE #580: Direct use of collections breaks with Python3.10
Various collections.* -> collections.abc.* aliases were removed in Python 3.10

This causes in libadalang tests like:

```
_Output:
_Traceback (most recent call last):
_  File "C:\libadalang\tmp\python__R918-040_rebindings\test.py", line 19, in <module>
_    ps = u.root.find(lambda n: n.text == 'Foo').parent
_  File "C:\libadalang\build\python\libadalang\__init__.py", line 2941, in find
_    return next(self.finditer(ast_type_or_pred, **kwargs))
_  File "C:\libadalang\build\python\libadalang\__init__.py", line 2972, in finditer
_    elif isinstance(ast_type_or_pred, collections.Sequence):
_AttributeError: module 'collections' has no attribute 'Sequence'
```

https://github.com/AdaCore/langkit/blob/28010dfe1c91d970991094aba8697e5b1e75bf34/langkit/templates/python_api/module_py.mako#L1506

should be

```python
        elif isinstance(ast_type_or_pred, collections.abc.Sequence):
```

(no change of import necessary, I believe)

LordAro:
Alias has been in place since Py 3.3, and minimum (test) version is 3.8, so I think we're safe

Figured I'd save everyone some time and PR it myself

Closes #580

(I'd have done 579 as well, but need your sign off on my proposed ~hack~workaround to do it)

### DIFF
--- a/langkit/templates/python_api/module_py.mako
+++ b/langkit/templates/python_api/module_py.mako
@@ -1503,7 +1503,7 @@ class ${root_astnode_name}:
         if isinstance(ast_type_or_pred, type):
             sought_type = ast_type_or_pred
             pred = lambda node: isinstance(node, sought_type)
-        elif isinstance(ast_type_or_pred, collections.Sequence):
+        elif isinstance(ast_type_or_pred, collections.abc.Sequence):
             sought_types = ast_type_or_pred
             pred = lambda node: isinstance(node, tuple(sought_types))
         else:


pmderodat:
Hello @LordAro,

Thank you for your contribution! Your patch looks good to me; before we merge it, could you replace “3.8” with “3.7” in the commit message (we target 3.7+) and wrap that message to fit in 76 columns?

LordAro:
Bleh, you made me actually check out the branch :p Done.

(If you squash the PR, GH UI allows you to reword commits)

pmderodat:
Ah, I did not realize that, but I prefered to check with you first anyway. Thanks!

pmderodat (approved):

PR #583, opened 2021-11-12
Lkt: add more PropertyError constraints

thvnx:
This patch allows to forbid the use of PropertyError as a type,
and enforces that ``raise`` can only calls a ``PropertyError()``.

### DIFF
--- a/contrib/lkt/language/parser.py
+++ b/contrib/lkt/language/parser.py
@@ -2285,8 +2285,17 @@ class SimpleTypeRef(TypeRef):
 
         return d.result_ref.then(
             lambda d: d.match(
-                # The type ref references a type decl: return no error
-                lambda _=T.TypeDecl: No(T.SemanticResult.array),
+                # The type ref references a type decl: return no error if it
+                # doesn't references a PropertyError.
+                lambda td=T.TypeDecl: If(
+                    td == Entity.property_error_type,
+                    [Entity.error(
+                        S("`").concat(Entity.property_error_type.full_name)
+                        .concat(S("` can only be referenced"))
+                        .concat(S(" in a raise expression"))
+                    )],
+                    No(T.SemanticResult.array)
+                ),
 
                 # Not a type decl: return an error that the type reference is
                 # invalid.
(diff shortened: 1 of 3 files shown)


contrib/lkt/language/parser.py
```diff
@@ -2582,19 +2590,34 @@ def check_correctness_pre():
                     )
                 )
             )
-            # In case of PropertyError, check for RaiseExpr
+            # In case of PropertyError, check presence of RaiseExpr
+            # and correct call to PropertyError exception.
```
pmderodat, line 2594: The comment is not very clear:

> […] check that this call is the operand of a raise expression and this is a call to the PropertyError constructor.

contrib/lkt/language/parser.py
```diff
@@ -2582,19 +2590,34 @@ def check_correctness_pre():
...
+                    Entity.expr_context_free_type
+                    == Entity.property_error_type,
 
-                    Self.error(
-                        S("cannot call ")
-                        .concat(Entity.property_error_type.full_name)
-                        .concat(S(" outside of a raise expression"))
-                    ).singleton,
+                    Cond(
+                        Not(Entity.first_no_paren_parent.is_a(T.RaiseExpr)),
```
pmderodat, line 2601: As we don’t have keyword to make it clear what is a condition and what is an expression for the result, we customarily group corresponding conditions/results together. In a conventional language, where we would have:

```
if C1:
   return E1
elif C2:
   return E2
else:
   return E3
```

we do this in the DSL:

```
Cond(
    C1,
    E1,

    C2,
    E2,

    E3,
)
```

Can you adjust the empty lines accordingly?

contrib/lkt/language/parser.py
```diff
@@ -2582,19 +2590,34 @@ def check_correctness_pre():
...
+
+                        Self.error(
+                            S("cannot call `")
+                            .concat(Entity.property_error_type.full_name)
+                            .concat(S("` outside of a raise expression"))
+                        ).singleton,
+
+                        Not(Entity.called_decl == Entity.property_error_type),
```
pmderodat, line 2609: Can you use the `!=` operator to make this condition easier to read?

contrib/lkt/language/parser.py
```diff
@@ -2285,8 +2285,16 @@ def check_correctness_pre():
...
             lambda d: d.match(
-                # The type ref references a type decl: return no error
-                lambda _=T.TypeDecl: No(T.SemanticResult.array),
+                # The type ref references a type decl: return no error if it
+                # doesn't references a PropertyError.
+                lambda td=T.TypeDecl: If(
+                    td == Entity.property_error_type,
+                    [Entity.error(
+                        S("`").concat(Entity.property_error_type.full_name)
+                        .concat(S("` is not a type"))
```
pmderodat, line 2294: I think we should avoid pretending that `PropertyError` is not a type (it’s a struct in the prelude, so technically it is a type). What about:
> PropertyError can only be referenced in a raise expression

pmderodat (changes requested):

thvnx:
@pmderodat, thanks for your comments and suggestions. Should be fixed accordingly now.

pmderodat (approved):

PR #585, opened 2021-11-16
support: allow override of ADAFLAGS, LDFLAGS, library soversion

asarhaddon:
All this would be useful when packaging the support library for Debian.
I have tried to copy the code from gnatcoll so that it also benefits windows and macos.
Thanks for considering.


support/langkit_support.gpr
```diff
@@ -12,6 +12,14 @@ library project Langkit_Support is
    Library_Kind_Param : Library_Kind_Type := external
      ("LIBRARY_TYPE", external ("LANGKIT_SUPPORT_LIBRARY_TYPE", "static"));
 
+   type OS_Kind is ("windows", "unix", "osx");
+   OS : OS_Kind := External ("LANGKIT_SUPPORT_OS", "windows");
```
pmderodat, line 16: We would like to keep things working with default options: please replace `"windows"` with `""` here and adjust the rest of the patch to avoid defining `Library_Version` in that case?

support/langkit_support.gpr
```diff
@@ -70,20 +78,35 @@ library project Langkit_Support is
...
+   case OS is
+      when "windows" => So_Ext := "dll";
+      when "osx"     => So_Ext := "dylib";
+      when "unix"    => So_Ext := "so";
+   end case;
+
+   case Library_Kind_Param is
+      when "relocatable" =>
+         for Library_version use "lib" & project'Library_Name & "." & So_Ext
+                                 & "." & Soversion;
```
pmderodat, line 91: Please fix the casing `Library_Version`

support/langkit_support.gpr
```diff
@@ -12,6 +12,14 @@ library project Langkit_Support is
    Library_Kind_Param : Library_Kind_Type := external
      ("LIBRARY_TYPE", external ("LANGKIT_SUPPORT_LIBRARY_TYPE", "static"));
 
+   type OS_Kind is ("windows", "unix", "osx");
+   OS : OS_Kind := External ("LANGKIT_SUPPORT_OS", "windows");
+
+   Soversion := External ("LANGKIT_SUPPORT_SOVERSION", "1");
```
pmderodat, line 18: For consistency inside the project file, could you rename `Soversion` to `SO_Version`, `Adaflags` to `Ada_Flags`, `Ldflags` to `Ld_Flags` and finally `So_Ext` to `SO_Ext`?

pmderodat:
Thank you for your contributions! In order to merge them, we need you to sign the CLA.

Regarding the commit messages: could you keep them wrapped at 76 columns?

pmderodat (changes requested):

raph-amiard:
@asarhaddon ping! We would love to merge this, but you need to answer @pmderodat 's last comment

asarhaddon:
Hello.
I had a more simple idea meanwhile.
Instead of concatenating a suffix depending on LANGKIT_SUPPORT_OS and LANGKIT_SUPPORT_SOVERSION into Library_Version, we should maybe support a single LANGKIT_SONAME variable.
If the variable is unset, Library_Version remains unset in order to keep the existing behaviour. If it is set, the content is assigned to Library_Version.
Both usage and implementation would be simplified.
Should I implement this idea or an "unknown" choice for LANGKIT_SUPPORT_OS?

pmderodat:
> […] we should maybe support a single LANGKIT_SONAME variable. If the variable is unset, Library_Version remains unset in order to keep the existing behaviour. If it is set, the content is assigned to Library_Version

This looks like a good idea. Yes, let’s do it, thank you!

PR #587, opened 2021-11-16
Remove dependency of lexer/regexp.py to lexer/unicode_data.py

asarhaddon:
The two removed lines seem to have no effect, and break the refresh of
unicode_data.py.


pmderodat (approved):
I wrote this code several years ago, but I don’t understand today why I’ve put these statements here. Thank you for the cleanup!

PR #589, opened 2021-11-16
support: remove Refcounted packages

asarhaddon:
These packages are not referenced within this library according to

They are not listed in the project Library_Interface, so they are
neither referenced from the outside nor compiled.


pmderodat (approved):
Good finding, thank you!

PR #599, opened 2021-11-26
Property DSL: fix handling of index in filter/take_while lambdas

pmderodat:


pmderodat:
This commit was mistakenly left in the #600 PR, so it is already merged.

PR #602, opened 2021-11-30
Lkt: complete typing of "if" expressions

pmderodat:


thvnx (approved):
Looks good!

contrib/lkt/language/parser.py
```diff
@@ -763,17 +779,27 @@ def expected_type():
...
 
             lambda if_expr=T.IfExpr: If(
                 # If we're resolving the type of the condition expression, then
                 # the result type needs to be a boolean.
                 Entity == if_expr.cond_expr,
                 Self.bool_type,
 
-                # Else, we're in a branch, use the expected type
-                if_expr.expected_type
+                # Else, we're in a branch: use type info from teh whole "if"
```
raph-amiard, line 800: Typo "teh"

contrib/lkt/language/parser.py
```diff
@@ -3152,6 +3178,94 @@ class IfExpr(Expr):
     alternatives = Field(type=T.ElsifBranch.list)
     else_expr = Field(type=T.Expr)
 
+    @langkit_property(return_type=T.Expr.entity.array)
+    def cond_exprs():
+        """
+        All expression that compute a condition.
```
raph-amiard, line 3184: All expression*s*

contrib/lkt/language/parser.py
```diff
@@ -3152,6 +3178,94 @@ class IfExpr(Expr):
...
+        """
+        return (
+            Entity.cond_expr.singleton
+            .concat(Entity.alternatives.map(lambda alt: alt.cond_expr))
+        )
+
+    @langkit_property(return_type=T.Expr.entity.array)
+    def result_exprs():
+        """
+        All expression that compute a result.
```
raph-amiard, line 3194: Same

contrib/lkt/language/parser.py
```diff
@@ -3152,6 +3178,94 @@ class IfExpr(Expr):
...
+        """
+        return (
+            Entity.then_expr.singleton
+            .concat(Entity.alternatives.map(lambda alt: alt.then_expr))
+            .concat(Entity.else_expr.singleton)
+        )
+
+    @langkit_property()
+    def check_correctness_pre():
+        # Check that all expression used in conditions return bools
```
raph-amiard, line 3204: Same

contrib/lkt/language/parser.py
```diff
@@ -3152,6 +3178,94 @@ class IfExpr(Expr):
...
+        invalid_conds = Entity.cond_exprs.map(
```
raph-amiard, line 3205: For the bool check, you're not supposed to need such a check here: The type errors should be sufficient, here you're duplicating the type checker's job in check correctness pre. 

For example, for the already existing check on the bool cond of the if branch, for this code:

```
@invalid val a = if 2 then 3 else 4
```

I get 

```
test.lkt:1:21: error: Mismatched types: expected `Bool`, got a number literal
1 | @invalid val a = if 2 then 3 else 4
  |                     ^
```

For the "alternatives return compatible types", I'm not sure exactly what's going on but I'm almost certain this shouldn't be a "pre" check if it needs to be a check.
Is  there a specific reason you added those checks here ?

raph-amiard: On second thought, the "alternative return compatible types" should be handled in `expected_type`, in a similar way to what is done for list elements. I tested this locally and it seems to work:

```
diff --git a/contrib/lkt/language/parser.py b/contrib/lkt/language/parser.py
index 50c974c51..af4425e9f 100644
--- a/contrib/lkt/language/parser.py
+++ b/contrib/lkt/language/parser.py
@@ -762,7 +762,11 @@ class Expr(LktNode):
                 Self.bool_type,

                 # Else, we're in a branch, use the expected type
-                if_expr.expected_type
+                if_expr.expected_type._or(
+                    # If there is no expected type, use the cftype of the then
+                    # branch.
+                    if_expr.then_expr.expr_context_free_type
+                )
             ),

             lambda err_on_null=T.ErrorOnNull: err_on_null.expected_type,
```

raph-amiard (changes requested):

contrib/lkt/language/parser.py
pmderodat: Ah right, I did not realize that this was duplicated work. TBH, the separation of work between the regular expr type checking framework and `check_correctness_pre` isn’t very clear to me yet. I’ll adjust, thanks.

pmderodat: Your tentative patch is an improvement, but I think we should do even better: regardless of the branch from which we call `expected_type`, we should try to use the context free type of all other branches. This is necessary to correctly infer types in cases such as:

```
val v = if true then [] elif false then [i] else []
```

contrib/lkt/language/parser.py
```diff
@@ -3241,6 +3251,42 @@ class IfExpr(Expr):
...
+        )
+
+    @langkit_property(return_type=T.TypeDecl.entity)
+    def if_expected_type():
+        """
+        Helper to handle ``if`` expressions in ``Expr.expected_type``.
+        """
+        # Use the if expression context free type if available, otherwise
+        # fallback on its expected type.
+        return Entity.expr_context_free_type.then(
```
raph-amiard, line 3285: you can use `._or` here

raph-amiard (changes requested):

contrib/lkt/language/parser.py
pmderodat: I didn’t know about it, thanks! Done, now.

raph-amiard (approved):

PR #603, opened 2021-11-30
Lkt lowering: enhance big integers handling

pmderodat:


testsuite/tests/contrib/lkt_semantic/array_type_concrete/test.out
```diff
@@ -407,19 +407,16 @@ Expr <StringLit test.lkt:31:29-31:32>
...
 
-test.lkt:31:35: error: Mismatched types: expected `Array[String]`, got an array
+test.lkt:31:35: error: Mismatched types: expected `Array[String]`, got `Array[Int]`
 31 | @invalid val a11 = [[bar(), "2"], [1, 2]]
    |                                   ^^^^^^
 
-test.lkt:31:20: error: ambiguous type for expression
-31 | @invalid val a11 = [[bar(), "2"], [1, 2]]
-   |                    ^^^^^^^^^^^^^^^^^^^^^^
+Expr <ArrayLiteral test.lkt:31:20-31:42>
```
thvnx, line 420: Shouldn't still be ambiguous here?

contrib/lkt/language/parser.py
```diff
@@ -129,6 +129,14 @@ class ParamMatch(Struct):
     formal = UserField(type=T.FormalParam, default_value=No(T.FormalParam))
```
thvnx, line 129: Comment about the commit itself. I guess it shouldn't be in this PR?

testsuite/tests/contrib/lkt_semantic/array_type_concrete/test.out
pmderodat: Why should it be? Note that we still have one error:
```
test.lkt:31:35: error: Mismatched types: expected `Array[String]`, got `Array[Int]`
31 | @invalid val a11 = [[bar(), "2"], [1, 2]]
   |                                   ^^^^^^
```
so in a way we’re doing a better error reporting job, now (less noise, still has the useful bit).

contrib/lkt/language/parser.py
pmderodat: I suppose you are referring to “Lkt: complete typing of "if" expressions” (27eb3ce). Yes indeed it should not be there, I’ll remove it, thank you!

testsuite/tests/contrib/lkt_semantic/array_type_concrete/test.out
thvnx: My bad, didn't see that!

thvnx (approved):
LGTM

raph-amiard (approved):

PR #614, opened 2021-12-08
Use gnatcoll_gmp static library

atupone:
If gnatcoll_gmp has been built supporting static & relocatable libraries, it uses relocatable, and test fails when mixing static and relocatable libraries

### DIFF
--- a/langkit/libmanage.py
+++ b/langkit/libmanage.py
@@ -897,6 +897,7 @@ class ManageScript:
         result = ['-XBUILD_MODE={}'.format(self.build_mode),
                   '-XLIBRARY_TYPE={}'.format(library_type),
                   '-XGPR_BUILD={}'.format(library_type),
+                  '-XGNATCOLL_GMP_BUILD={}'.format(library_type),
                   '-XXMLADA_BUILD={}'.format(library_type)]
 
         if self.enable_build_warnings:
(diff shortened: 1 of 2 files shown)


atupone:
There is still other similar problem for gprbuild not telling about using relocatable or static libraries to this other:

- testsuite/tests/langkit_support
- testsuite/tests/adalog
- testsuite/tests/misc/link_two_libs

pmderodat:
Hello,

We can’t reproduce any problem like that. The project file for gnatcoll-gmp from sources (https://github.com/AdaCore/gnatcoll-bindings/blob/master/gmp/gnatcoll_gmp.gpr#L36) uses the `LIBRARY_TYPE` variable name, and so does the installed project (https://github.com/AdaCore/gnatcoll-bindings/blob/master/gmp/setup.py#L56), so the current GPRbuild arguments passing should be fine. Actually, nowadays both XML/Ada and LibGPR use the `LIBRARY_TYPE` name, so we could get rid of both the `GPR_BUILD` and `XMLADA_BUILD` ones.

Could you clarify where the `GNATCOLL_GMP_BUILD` name comes from in your setup?

atupone:
OK, I had to specify --build-var=LIBRARY_TYPE to gprinstall gnatcoll-bindings. 
GNATCOLL_GMP_BUILD was just the default as in the manual for build-var :
**_Specify the name of the build variable in the installed project. If this options is not used, the default build variable used is <PROJECT_NAME>_BUILD._**

No need to apply this pull request, so. But need to take care of the others as the problem still exist for the gprbuild call that do not specify the LIBRARY_TYPE to use.

PR #617, opened 2021-12-13
New solver

pmderodat:


support/langkit_support-adalog-solver.ads
```diff
@@ -45,8 +45,9 @@ package Langkit_Support.Adalog.Solver is
    --  depth. Solving a relation will assign values to every logic variable
    --  involved in the relation or fail.
    --
-   --  A relation is manually ref-counted, and owns every sub-relation. When
-   --  the ref-count reaches 0, every sub-relation is destroyed.
+   --  A relation is manually ref-counted, and has an ownership share of every
+   --  sub-relation. When the ref-count of a root relation, reaches 0, the
```
raph-amiard, line 57: Weirdly placed ","

Roldak (commented):
LGTM. The only comment I have is about the equation pretty-printing. We currently display:
```
42 -> %X
```
Whereas with the old solver we were printing
```
%X <- 42
```
And IMO it's easier to read.

support/langkit_support-adalog-generic_main_support.adb
```diff
@@ -88,10 +88,21 @@ package body Langkit_Support.Adalog.Generic_Main_Support is
...
+
+      case Kind is
+      when Symbolic =>
+         Put_Line ("... without optimizations:");
+         Solve (Rel, Solution_Callback'Access, (Cut_Dead_Branches => False));
+         New_Line;
+
+         Put_Line ("... cut dead branches:");
+         Solve (Rel, Solution_Callback'Access, (Cut_Dead_Branches => True));
+         New_Line;
```
raph-amiard, line 100: It seems that it would be good to at least check and show an "ERROR" or "WARNING" when we have diverging solutions in both cases, because for now it would be easy to miss it. WDYT ?

pmderodat: Will do, thanks! We’ll change it to a real error (not one that can be baselined) once the current aliasing bug is fixed.

raph-amiard (approved):

PR #619, opened 2021-12-15
Rework LKT check semantic pass

raph-amiard:


contrib/lkt/language/parser.py
```diff
@@ -1191,110 +1191,68 @@ def lexer():
     )
 
     @langkit_property()
-    def check_correctness_pre():
-        # All grammars must have an associated lexer. First check that the
-        # @with_lexer annotation is present.
+    def check_legality():
```
Roldak, line 1194: It seems to me that in order to be able to resolve names inside a grammar definition, we need to be sure that the lexer has been correctly resolved first (because there can be references to tokens inside the grammar). Since this check is now done as a legality check, it seems to me that it is too late and some references in the grammar will fail just because the referenced lexer couldn't be resolved.

Roldak (changes requested):

contrib/lkt/language/parser.py
raph-amiard: You're right, good catch. I'll put back some or all of those checks in `check_pre`

contrib/lkt/extensions/mains/lkt_toolbox.adb
```diff
@@ -41,6 +41,13 @@ procedure Lkt_Toolbox is
          Long   => "--check-only",
          Help   => "Only output the errors");
 
+      package Flag_Invalid is new Parse_Flag
+        (Parser => Parser,
+         Short  => "-I",
+         Long   => "--flag-invalid-nodes",
+         Help   => "Flag nodes that generate errors that are not annotated"
+                   & " with the @invalid annotation. Also flag nodes "
```
pmderodat, line 49: For consistency and easy check of missing spaces, could you move the the trailing space to the beginning of next line?

```
& "... Also flag nodes"
& " annotated with …"
```

Also given that `@invalid` is for decls and that we check both decls with and without `@invalid`, I would suggest to call this option `--check-invalid-decls`.

contrib/lkt/language/parser.py
```diff
@@ -422,19 +422,23 @@ def check_sem_internal(
     @langkit_property(return_type=T.SemanticResult.array, memoized=True)
     def check_correctness_post():
         """
-        Custom hook to implement legality checks for a given node that can
+        Custom hook to implement checks for a given node that can
```
pmderodat: Why do we have only “checks” here, and “correctness” checks below? What’s the difference between “legality checks” and “correctness checks” anyway? An explanation somewhere would help.

contrib/lkt/language/parser.py
```diff
@@ -422,19 +422,23 @@ def check_sem_internal(
...
-        Custom hook to implement legality checks for a given node that need to
-        run *before* type resolution. If no errors, returns a null array.
+        Custom hook to implement correctness checks for a given node that need
+        to run *before* type resolution. If no errors, returns a null array.
 
         .. WARNING: This must *not* raise exceptions, so must make sure that
             any queried semantic results are queried in a safe fashion.
+
+            In general **ONLY USE THIS FOR CHECKS THAT SHOULD RUN BEFORE NAME
+            AND TYPE RESOLUTION**, ie. checks that if they're not made will
```
pmderodat, line 427: “i.e.”

contrib/lkt/language/parser.py
```diff
@@ -332,42 +350,15 @@ def type_mismatch_error(expected=T.TypeDecl.entity, got=T.TypeDecl.entity):
...
-
-        has_internal_errors = Var(errs.any(lambda r: r.error_message.is_null))
-
-        return If(
-            has_internal_errors & (diags.length == 0),
-            PropertyError(T.TreeSemanticResult,
-                          "ERROR: internal errors without diagnostics"),
-            res
+    @langkit_property(return_type=T.SemanticResult.array)
+    def check_legality_impl():
```
pmderodat, line 354: Can you add a docstring for this? `check_legality` (at least defined line 403) does not call this, which is the usual pattern with `_impl` properties (here it’s the other way around)

pmderodat (changes requested):

contrib/lkt/language/parser.py
```diff
@@ -1192,109 +1192,69 @@ def lexer():
...
+                .singleton
             )
-        )
 
-        # Now check the grammar rules themselves
-        rules = Var(Entity.rules.filter(
-            lambda d: d.decl.is_a(T.GrammarRuleDecl)
-        ))
+            # Then check that the associated arguments are correct
+            .concat(((lexer_annot_params.children.length != 1)
```
pmderodat, line 1216: This crashes when `lexer_annot_params` is null (previous code checked `Not(params.is_null)`)

contrib/lkt/language/parser.py
```diff
@@ -1843,7 +1843,10 @@ def is_authorized_decl(decl=T.Decl.entity):
         pass
 
     @langkit_property()
-    def check_correctness_pre():
+    def check_legality():
+        # TODO: This implementation implies that we'll run full checking for
+        # all decls, even if they're in a context where we don't want to have
+        # them. Is that a good idea?
```
pmderodat, line 1849: I don’t understand this: if I understand correctly the new organization for checks, `check_correctness_pre` is the called for all nodes while `check_legality` is called for nodes in a decl only when only when name resolution/typing validated that decl. What context are you referring to? Also can you expand the commit message to explain what’s the purpose of moving the checks?

testsuite/tests/contrib/lkt_semantic/decl_block_correctness/test.out
```diff
@@ -1,25 +1,5 @@
 Resolving test.lkt
 ==================
-test.lkt:5:4: error: value declaration forbidden in trait declaration
```
pmderodat, line 3: I realize now that we have error messages in a disturbing order. Would it be possible to enhance either a property or lkt_toolbox to sort error messages? Test updates would be much easier to review, and I think it would be user friendlier for users too. WDYT?

pmderodat (changes requested):

contrib/lkt/language/parser.py
raph-amiard: This comment is completely reworked later in the commit chain, so this is just a bleep. Will get rid of it

contrib/lkt/language/parser.py
```diff
@@ -1473,6 +1462,27 @@ def ref_not_found_error():
...
+
+        return If(
+            # Check that we don't reference the error type (PropertyError)
+            # outside of a raise expression
+            # NOTE: This means that we lookup the property error type for every
+            # RefId in the code. That might be too costly, so we might want to
+            # have a first level of textual filter if this ever becomes a
+            # problem.
+            And(Entity.referenced_decl.result_ref == error_type,
+                Not(Entity.parents.any(lambda r: r.is_a(T.RaiseExpr)))),
```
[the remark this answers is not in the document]
raph-amiard: Actually your message made me realize that it's worse than that, since currently you can only pass a static string in Langkit to the `PropertyError` constructor.  I'll make the check stricter in `RaiseExpr.check_legality`, which should also take care of the case you're mentioning.

contrib/lkt/language/parser.py
```diff
@@ -3348,21 +3314,22 @@ class RaiseExpr(Expr):
...
+            ),
 
-            Self.except_expr.error(
-                S("raised expression needs to be of type `").concat(
-                    pe_type.full_name
-                ).concat(S("`, got `")).concat(
-                    except_expr_type.full_name
-                ).concat(S("`"))
-            ).singleton
+            default_val=error_msg
```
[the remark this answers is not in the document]
raph-amiard: This was refactored as per the suggestion above, so not applicable anymore at least in this form

contrib/lkt/language/parser.py
raph-amiard: This comment is in the implementation of whether a decl is authorized in a certain context or not. Since we do that as a semantic check, and do not exempt further resolution, I added this comment because I think it might happen that having a declaration in a context where it is not authorized might generate further message, or in other words, we might want to exempt the whole decl from further messages.

I moved it because so far it has no functional reason to be in `check_correctness_pre`, and the whole point of this refactoring was to have nothing in there that you can avoid having there. I think this is plenty explained in comments in this PR.

testsuite/tests/contrib/lkt_semantic/decl_block_correctness/test.out
raph-amiard: I'm not against it at first sight, but won't do this as part of this PR, since this was already the case before, and I need to think on it a bit to see if it doesn't have any hidden drawbacks

contrib/lkt/language/parser.py
pmderodat: Ok thanks

pmderodat (changes requested):

contrib/lkt/language/parser.py
```diff
@@ -3290,15 +3291,16 @@ def check_legality():
...
-            default_val=error_msg
-        )
+            # The called declaration is the PropertyError type
+            ce.name.check_referenced_decl != error_type,
+
+            # If the call has an argument it is a static string expression
+            Not(ce.match_params().at(0).then(
+                lambda p: p.actual.value.is_a(StringLit), default_val=True))
+
+        ).then(lambda _: error_msg), default_val=error_msg)
```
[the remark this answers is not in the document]
raph-amiard: I don't find it confusing ... I can fix the comments if you want ?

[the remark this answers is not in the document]
raph-amiard: I put your version in because whatever

pmderodat (approved):

Roldak (approved):

PR #621, opened 2022-01-13
Remove useless recursions in parsers handling

pmderodat:


thvnx:
Definitively solves my issue, thanks!

PR #626, opened 2022-01-27
dsl_unparse: print diagnostics when getting analysis unit fails

thvnx:
Print parsing diagnostics in case getting analysis unit fails in order
to help developers to spot the error(s). The print method introduced
here comes from kalint.py tutorial file.

TN: V127-026


langkit/dsl_unparse.py
```diff
@@ -20,6 +22,17 @@
 def fqn(prop):
     return "{}.{}".format(prop.struct.name.camel, prop._original_name.lower)
 
+def print_diagnostic(filename, sloc_range, message):
+    line = sloc_range.start.line
+    column = sloc_range.start.column
+    print('In {}, line {}:'.format(filename, line), file=sys.stderr)
+    with open(filename) as f:
+        # Get the corresponding line in the source file and display it
```
raph-amiard, line 30: You should probably use `langkit.diagnostics` here, rt. reinventing the wheel.

raph-amiard (changes requested):

langkit/dsl_unparse.py
thvnx: Done ;). For info, output looks like:

```
ast.py:9174:13: error: Expected ')', got Indent
9173 |             ).map(
     |                   
```

thvnx:
Obsolete, moved to internal PR system.

PR #640, opened 2022-03-23
UA21-001 Don't increase/decrease logging indent...

godunko:
... when logging is not enabled. It modify indentation value of the
GNATCOLL.Traces's logging stream, thus unexpected (huge) indentation is
generated in multi thread applications.


pmderodat:
Thanks for the patch. We just merged it: https://github.com/AdaCore/langkit/pull/640/commits/3ef4d4309ca862e2c90c1f6a9041ab2ac7f7fe3b

PR #642, opened 2022-04-03
Stop depending on the funcy non-standard module

asarhaddon:
Each dependency outside the standard library slightly increases the
maintenance cost and the difficulty for new contributors.

Readability is always subjective, but performances are critical for
langkit and the changes introduced by this commit should sligthly
improve it:
* Several intermediate list constructions are replaced by integer
  list/tuple indices or Python generators.
* In langkit/expressions/base.py, traversal of a tree by a recursive
  generator is replaced with a single loop.  Moreover, according to
  https://github.com/Suor/funcy/blob/master/funcy/tree.py, the funcy
  implementation uses a deque structure while only right-appends are
  actually needed.

The only dubious change in langkit/dsl_unparse.py constructs an
intermediate tuple, but the previous version was concatenating
intermediate lists, which is slightly worst.

### DIFF
--- a/REQUIREMENTS.dev
+++ b/REQUIREMENTS.dev
@@ -3,7 +3,6 @@ coverage==5.1
 docutils==0.16
 -e git+https://github.com/AdaCore/e3-core.git#egg=e3-core
 -e git+https://github.com/AdaCore/e3-testsuite.git#egg=e3-testsuite
-funcy==1.14
 Mako==1.1.2
 sphinx-rtd-theme==0.4.3
 yapf==0.30.0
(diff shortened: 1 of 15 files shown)


pmderodat:
Hello,

First of all, thank you for contributing to Langkit!

Even though as a general principle in software development, each new dependencies has the potential to make things harder for development setup and packaging, in the end it’s always a tradeoff: how hard it is to get this dependency (get the correct version, setup transitive dependencies, etc.) vs. how nice using it in the codebase is (less efforts, more readable code, etc.).

In the case of funcy, we feel that it’s worth using it: given that it is a pure Python library (no native extension or other odd setup), there should be no development setup hassle (especially with `pip` these days) and no packaging issue neither (for instance I see it available both in [Debian](https://packages.debian.org/sid/python3-funcy) and [ArchLinux](https://archlinux.org/packages/community/any/python-funcy/)).

For theses reasons, we’re sorry to say we won't merge this PR, even though we appreciate the work you put in it. For future occurences, what about first opening an issue to first discuss if/how to solve some problem?

PR #644, opened 2022-08-03
create-project.py: forward verbosity option

ronan-d:
Hi, this is a change that makes it easier to turn on verbose mode completely.

I also considered setting `self.context.verbosity` directly in `class ManageScript`, but I think that would have changed the API.

### DIFF
--- a/scripts/create-project.py
+++ b/scripts/create-project.py
@@ -55,7 +55,8 @@ class Manage(ManageScript):
 
         return CompileCtx(lang_name={lang_name_repr},
                           lexer={lang_name_slug}_lexer,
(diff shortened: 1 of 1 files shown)


pmderodat:
Hello @ronan-d,

Thank you for your contribution! Sorry for the time it took to get there, but your commit is now merged in the master branch: https://github.com/AdaCore/langkit/commit/fd7086d5765c5b431a083a125cb9deec5c0b024b Because of reasons I don’t control, it is not possible to merge GitHub PRs directly, so I have to manually close it, but in practice your change is merged.

PR #646, opened 2023-02-25
getargspec removed on python 3.11

atupone:


pmderodat:
Hello,

Thank you! We have merged your changes as https://github.com/AdaCore/langkit/commit/52084ad85d3c0179036c563f3e332cdcd003c236

Note that we had to tweak the patch a little, as the part in `expressions/base.py` contained `argspec.kw`, whereas `kw` is not a valid attribute (that was visible in the testsuite): the correct attribute name is `varkw`.

PR #652, opened 2024-01-29
setup.py: rename Windows executable from 'create-project.py.exe' to 'create-project.exe'

sad-poet:
It is not necessary to add the `.py` suffix to the script name of the console entry point.

For each `console_scripts` entry point in `setup.py`, setuptools generates a wrapper executable, which is not itself a Python script. For example, on Windows, it would produce a standard PE binary with name `create-project.py.exe` that calls the Python interpreter with the configured script.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ setup(
     ]},
     entry_points={
         "console_scripts": [
-            "create-project.py = langkit.scripts.create_project:main",
+            "create-project = langkit.scripts.create_project:main",
         ]
     },
 )


pmderodat:
Hello @sad-poet,

Thank you for your contribution! It has just been merged on the `master` branch: https://github.com/AdaCore/langkit/commit/369b9a89dae2060ac5d16dd85ca0905e16bba3bd

GitHub has not detected that we merged it using our internal repository management system, so closing the MR here.

